Issue #6, opened 2015-09-06
support/install_command.ps1 should be under 3040 with command variables and base64 encoded


hh:
I have a feeling that [support/install_command.ps1](https://github.com/chef/mixlib-install/blob/master/support/install_command.ps1) longer than what we've had in the past. WinRb/WinRM#41 notes that ```3040``` seems to be the max and I'm seeing ```3058``` now. If we can drop that by at least 18 characters or so I think things will work.

Errors are showing up in chef/knife-windows#262 and chef/chef-provisioning#431 when trying to hit windows2012 nodes on ec2.

Do we have a test suite for knife-ec2 and chef-provisioning-ec2 hitting  for ec2 windows2012 nodes?

https://github.com/chef/mixlib-install/commit/a83a59e47508e3e710e13a6e963c39472b1378a8#commitcomment-13086621

hh:
Maybe we could use Chef::Provisioning::Transport::WinRM#write_file (which chunks by default) to write the install command out to a file, then execute that via scheduled task.

patrick-wright:
Fixed in TK

PR #7, opened 2015-09-06
sizing of install_command.ps1 < 3040bytes when interpolated

hh:
These changes to human messages where the quickest way to reduce the size.
I'm sure someone proficient in PowerShell could do much better.

### DIFF
--- a/support/install_command.ps1
+++ b/support/install_command.ps1
@@ -37,18 +37,18 @@ Function Download-Chef($url, $md5, $dst) {
   } Finally { if ($c -ne $null) { $c.Dispose() } }
 
   if ($md5 -eq $null) { Log "Skipping md5 verification" }
-  elseif (($dmd5 = Get-MD5Sum $dst) -eq $md5) { Log "Successfully verified $dst" }
-  else { throw "MD5 for $dst $dmd5 does not match $md5" }
+  elseif (($dmd5 = Get-MD5Sum $dst) -eq $md5) { Log "Verified $dst" }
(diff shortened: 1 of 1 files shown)


hh:
Should fix chef/mixlib-install#6  chef/knife-windows#262 and chef/chef-provisioning#431

thommay:
I appreciate that this is a tricky and annoying problem, but I believe the right solution is to figure out a sustainable approach cross versions (this is broken on win2k8 for different reasons, involving memory limits on powershell sessions). To the best of my knowledge knife-windows doesn't (yet) use mixlib-install.

Also, to make this more annoying, the interpolated size is dependent on the string length of the tempdir, which we can't control.

I believe the correct approach is to use a scheduled task to perform the installation, which isn't subject to the string size limits, nor to the memory size problems.

hh:
A scheduled task would be interesting to monitor via a  single command, but I'm sure it can be done in under the max command size.

Until we get there, I think we should consider reducing the size of this file to fix at least a few use cases.

hh:
Maybe we could use [Chef::Provisioning::Transport::WinRM#write_file](https://github.com/chef/chef-provisioning/blob/4638bdf99e048452b5b80e06c118ba2a0a0132a3/lib/chef/provisioning/transport/winrm.rb#L55) (which chunks by default)  to write the install command out to a file, then execute that via scheduled task.

thommay:
Yes, that's the plan. I'm closing this PR since the work is mostly in chef/chef-provisioning#410

PR #10, opened 2015-10-01
[WIP] ScriptGenerator

patrick-wright:
Not in a working state.  PR created for review and discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,9 @@ Or install it yourself as:
 
 ## Usage
 
-TODO: Write usage instructions here
+```ruby
+Mixlib::Install::ScriptGenerator.new('latest')
+```
 
 ## Development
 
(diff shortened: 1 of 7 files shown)


sersut:
This is good. Exactly what I was thinking. 

- We should update the `README.md` with the instructions to use the `ScriptGenerator` class. 
- We should fix the tests so this PR is mergable.

patrick-wright:
@sersut Let me know whether this is on the right track.  I believe we are maintaining existing functionality while working in parallel so refactors can be done incrementally.  Should I be breaking `Mixlib::Install.install_command` at this point or should we allow a passthrough that calls `ScriptGenerator`'s `#install_command`?  I'm cool with breaking it in master though!

sersut:
I think for now we can remove `install_command` from the main class of `Mixlib::Install`. 

This is a breaking change and we will make sure kitchen and provisioning are well locked before release.

patrick-wright:
updated the PR change in PR 11.  Closing this PR.

PR #11, opened 2015-10-01
Omnitruck backend for mixlib-install

patrick-wright:
@sersut
branched from pw/script_gen


README.md
```diff
@@ -22,7 +22,9 @@ Or install it yourself as:
 
 ## Usage
 
-TODO: Write usage instructions here
+```ruby
+Mixlib::Install::ScriptGenerator.new('latest')
```
sersut: You need to call `install_command` on it no?

lib/mixlib/install/exception.rb
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+module Mixlib
+  class Install
+    class Exception
+      class UnsupportedChannel < ArgumentError
```
sersut: Rather than having a dedicated file for an exception class, we should collect them either in a single file or wherever they are relevant. We can have this class defined in the `Base` for now.

sersut:
Other than a couple of minor things :+1:

patrick-wright:
@thommay can we get your input before merging this breaking change?

.travis.yml
```diff
@@ -3,5 +3,4 @@ sudo: false
 rvm:
   - 2.2
   - 2.1
-  - 1.9.3
```
patrick-wright: @thommay are you okay with removing 1.9.3 testing?  We would like to use ruby 2.x features.

thommay: Nope. Test Kitchen still supports 1.9 and will do until it has a 3.0 release.

thommay:
So, I'm :-1: on this because we need to support 1.9.3 still.

.travis.yml
patrick-wright: Okay.  I'll make the back compat change today.

patrick-wright:
@thommay @sersut changed constructor to an options model.  1.9.3 ran and passed in CI.

thommay:
Please check what the version pins on test-kitchen and provisioning are like for this project, then feel free to merge.

patrick-wright:
@sersut I've added the Omnitruck backend implementation with specs.  
* The Install constructor now takes an options hash which get set in an Options class.  I have also moved the options validation here.
* Replaced rest-client with good ol' fashioned net http.
* Add self.from_json and to_hash methods to ArtifactInfo

patrick-wright:
kitchen 1.4.0 pins rubies:
```
rvm:
  - 2.2
  - 2.1
  - 2.0.0
  - 1.9.3
  - ruby-head
```

however, master now looks like:
```
rvm:
  - 2.2.3
  - 2.1
  - 2.0.0
  - ruby-head
```

provisioning only pins 
```
rvm:
  - 2.1
```

I would say it's safe to remove 1.9.3, but if we want to wait for an official release I will create a chore to do so.

lib/mixlib/install/options.rb
```diff
@@ -0,0 +1,36 @@
+module Mixlib
+  class Install
+    module Options
```
sersut: Thinking about this again, I think we should make this a class with some readers to get the option values. It should validate the parameters in its constructor. 

Previously I was thinking we can mixin this into the classes we want to support the options. But thinking about the `self.info` method above, this will not be really convenient since class methods will not have access to options via the mixed helpers.

lib/mixlib/install/backend.rb
```diff
@@ -0,0 +1,51 @@
...
+#
+
+require 'net/http'
+require 'json'
+
+module Mixlib
+  class Install
+    class Backend
+      def self.info(options)
+        parameters = {
```
sersut: Here I would do something like:

```
backend = if ['current', 'stable'].include? options.channel
  Backend::Omnitruck.new(options)
end

backend.info
```

And move the rest of the logic into Omnitruck class.

spec/spec_helper.rb
```diff
@@ -1,2 +1,5 @@
-$LOAD_PATH.unshift File.expand_path("../../lib", __FILE__)
-require "mixlib/install"
+$LOAD_PATH.unshift File.expand_path('../../lib', __FILE__)
+require 'mixlib/install'
```
sersut: I would not include :allthethings: here in spec_helper. Even though this makes writing specs a little bit more convenient, it sometimes masks require bugs. E.g. `mixlib/install' might not require `mixlib/artifact_info` even though it needs it.

spec/mixlib/install/backend_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+              url: 'http://opscode-omnibus-packages-current.s3.amazonaws.com/mac_os_x/10.9/x86_64/chef-12',
+              version: '12'
+            }
+          }
+
+          it_behaves_like 'the right artifact info'
+        end
+
+        context 'with latest version keyword' do
+          let(:project_version) { 'latest' }
```
sersut: IMO since we support symbols in `channel` a user would probably expect us to support `:latest` in the version as well.

sersut:
Let's sync back on the 1.9 support tomorrow @patrick-wright. If that is going away for kitchen we can totally make that go away too. 

Also let's check how `kitchen` and `provisioning` is pinning `mixlib-install`. We want to make sure that they will not get `0.8.0` or `1.0.0` when this code gets released before we change them.

thommay:
1.9.3 is not going away until 3.0 : https://github.com/test-kitchen/test-kitchen/pull/804#issuecomment-132788219

sersut:
@thommay I think this is ready to do. We have also checked kitchen and provisioning and their version pinning looks right:

https://github.com/test-kitchen/test-kitchen/blob/master/test-kitchen.gemspec#L31
https://github.com/chef/chef-provisioning/blob/master/chef-provisioning.gemspec#L21

patrick-wright:
Want to bump the version to 0.8.0?

sersut:
We will need to bump it to 1.0 since we have a breaking change. But we are not going to release this to anywhere yet so I think not bumping is fine. Unless @thommay has other things in mind that is :smile:

patrick-wright:
I think we should still bump the version to 0.8 since we are still in development mode (0.y).  We have some liberties even though this is a breaking change.  Breaking changes really apply to the first 1.0 "public" release and beyond.

thommay:
I think this should be 0.8, but I also don't think the version pins in tk or provisioning are going to stop this being installed: `~> 0.7` means `>= 0.7 < 1.0`. So you'll need to update them first.

thommay:
I'm :-1: on removing finstyle.

sersut:
What is the purpose of the finstyle in this context? We are using rubocop directly in other projects and didn't run into issues I believe.

thommay:
It has a sensible set of defaults. I'm completely :-1: on PRs that change things like single to double quotes whilst we're doing other things, too.

sersut:
K. I will:

- Remove rubocop related things. 
- Pin kitchen and provisioning correctly. 
- Bump version to 0.8.0. 

Anything else?

thommay:
Can you repush once the style changes are reverted please?

sersut:
Sure.

sersut:
Done.

lib/mixlib/install/backend.rb
```diff
@@ -0,0 +1,33 @@
...
+# limitations under the License.
+#
+
+require 'mixlib/install/backend/omnitruck'
+
+module Mixlib
+  class Install
+    class Backend
+      def self.info(options)
+        backend = if ['current', 'stable'].include? options.channel.to_s
```
thommay: I'm not really happy about this being hard coded but I don't think it's a blocker right now.

thommay:
I'm sorry to lean so heavily on the style stuff first up, but right now this PR generates:
```
16 files inspected, 132 offenses detected
```
with the preexisting rubocop config. Most of them look autocorrectable. 

The code looks fine; have you tested by dropping this in to provisioning or test kitchen?

sersut:
@thommay OK. I will fix them and hook it up so that we see the styling things in Travis. 

I am pretty sure they will break. I do not think that we need to update test-kitchen or provisioning until we have more complete functionality in here. Do you want us to update them now and update them again once we have the full functionality?

thommay:
No I don't want you to update them yet, I just want to make sure that you're testing this code in the wild somehow :)

sersut:
The new code we are adding is not usable by them yet. We will change test-kitchen in the context of this story: https://chefio.atlassian.net/browse/GS-3 which will be coming up next after GS-2.

lib/mixlib/install/backend/omnitruck.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        def info
+          parameters = {
+            p: options.platform,
+            pv: options.platform_version,
+            m: options.architecture,
+            v: options.product_version
+          }
+
+          endpoint = "https://omnitruck.chef.io"
```
schisamo: this should be a constant:

```ruby
ENDPOINT = 'https://omnitruck.chef.io'.freeze
```

We'll probably want to make that a configurable option (with this as a default) in the future.

patrick-wright: Agreed. We'll do that in this branch: https://github.com/chef/mixlib-install/compare/sersut/info-without-platform-details#diff-be25512b4fda9317c6049a36db712e99R27

lib/mixlib/install/script_generator.rb
```diff
@@ -0,0 +1,217 @@
...
+module Mixlib
+  class Install
+    class ScriptGenerator
+      attr_accessor :version
+
+      attr_accessor :powershell
+
+      attr_accessor :prerelease
+
+      attr_accessor :nightlies
```
schisamo: IMO we should not support the `:prerelease` and `:nightlies` options.

patrick-wright: This class used be in Mixlib::Install.  We just moved it as is to the
existing logic would still work.  This will be refactored to generate the
install scripts, and the old omnitruck configurations will go away.

lib/mixlib/install/script_generator.rb
```diff
@@ -0,0 +1,217 @@
...
+      #
+      # @param command [String] command to be prefixed
+      # @return [String] the command, conditionaly prefixed with sudo
+      # @api private
+      def sudo(script)
+        use_sudo ? "#{sudo_command} #{script}" : script
+      end
+    end
+  end
+end
```
schisamo: I would recommend making 3 classes here

* `Mixlib::Install::ScriptGenerator` (module only)
* `Mixlib::Install::ScriptGenerator::Base`
* `Mixlib::Install::ScriptGenerator::Bourne`
* `Mixlib::Install::ScriptGenerator::PowerShell`

The last two extend from `Mixlib::Install::ScriptGenerator::Base` which exposes numerous abstract methods that must be overridden. The module could expose a class method like `self.generator_class_for_options(opts)`. 

All of this should make things easier to reason about, test and maintain. 

Take a look at Omnibus's packagers for a decent working example:
https://github.com/chef/omnibus/blob/master/lib/omnibus/packager.rb
https://github.com/chef/omnibus/tree/master/lib/omnibus/packagers

I would also follow a similar pattern for the backends and any other objects that have `1..n` variants.

lib/mixlib/install/script_generator.rb
```diff
@@ -0,0 +1,217 @@
...
+        Util.shell_var(name, value, powershell)
+      end
+
+      # @return the correct Chef Omnitruck API metadata endpoint, based on project
+      def metadata_endpoint_from_project(project = nil)
+        if project.nil? || project.downcase == "chef"
+          "metadata"
+        else
+          "metadata-#{project.downcase}"
+        end
```
schisamo: This is all out of date now. You want to use: `/<CHANNEL>/<PROJECT>/metadata`

lib/mixlib/install/script_generator.rb
```diff
@@ -0,0 +1,217 @@
...
+      def windows_metadata_url
+        base = if omnibus_url =~ %r{/install.sh$}
+          "#{File.dirname(omnibus_url)}/"
+        end
+
+        url = "#{base}#{endpoint}"
+        url << "?p=windows&m=x86_64&pv=2008r2" # same package for all versions
+        url << "&v=#{CGI.escape(version.to_s.downcase)}"
+        url << "&prerelease=true" if prerelease
+        url << "&nightlies=true" if nightlies
```
schisamo: Similar to another comment...kill `prerelease` and `nightlies` options in all parts of this code base. Channels have fully replaced things.

lib/mixlib/install/script_generator.rb
```diff
@@ -0,0 +1,217 @@
...
+      def initialize(version, powershell = false, opts = {})
+        @version = version || "latest"
+        @powershell = powershell
+        @http_proxy = nil
+        @https_proxy = nil
+        @install_flags = nil
+        @prerelease = false
+        @nightlies = false
+        @endpoint = "metadata"
+        @omnibus_url = "https://www.chef.io/chef/install.sh"
```
schisamo: https://omnitruck.chef.io/install.sh

lib/mixlib/install/script_generator.rb
sersut: Totally agreed. This will come in a future PR when we are looking into ScriptGenerator. For now we just moved the functionality as is until we have all the underlying things ready.

sersut:
@schisamo thanks for the comments. We will visit ScriptGenerator in a future PR. Fir we want to make sure that all the underlying things are ready before switching to that.

sersut:
Just had a conversation with @schisamo and we will handle the `ScriptGenerator` comments in a separate PR. Shipping this in order to unblock our progress.

PR #14, opened 2015-10-09
Unstable channel support using Artifactory backend

patrick-wright:


rhass-r7:
This may be a silly question/distraction -- but how does this behave when artifactory is not available accessible. The merge of channels in the `ALL_SUPPORTED_CHANNELS` seems to assume it is always available - but I may be overlooking something.

lib/mixlib/install/backend/artifactory.rb
```diff
@@ -0,0 +1,65 @@
...
+  class Install
+    class Backend
+      class Artifactory
+        ARTIFACTORY_ENDPOINT = "http://artifactory.chef.co".freeze
+
+        attr_accessor :options
+
+        def initialize(options)
+          @options = options
+          @client = ::Artifactory::Client.new(endpoint: ARTIFACTORY_ENDPOINT)
```
sersut: an `attr_accessor` for `@client` please :smile:

patrick-wright: Just for sport? :)  Do we really need one for the client?

sersut:
Can we also raise a specific error that indicates that you can only use a specific version if you are using unstable channel?

patrick-wright:
@rhass-r7 Not a distraction.  In fact, you made a good point which would probably confuse people.  We are moving to a channel model (stable, current, etc).  Stable and current are publicly available via omnitruck.  We've introduced a new channel named unstable.  It is undecided if/when they will be made public.  The unstable channel uses Chef's internal artifact repository to query artifact metadata.

schisamo:
@rhass-r7 And just to be clear, most people will not want any software coming from the __unstable__ channel. We will be stashing software here that has been built but not validated with any sort of acceptance testing. The reason we want to teach mixlib-install how to work with the unstable channel is all of our acceptance testing tools will use mixlib-install to query/fetch/install packages from this channel.

sersut:
@patrick-wright can we add some instructions to the README about the fact that some specs require VPN access (access to internal Chef network) and here are the steps to run them. 

This is :+1: but let's wait for @thommay to chime in since we are pulling in a new dep and introducing some tests that can not be run via Travis.

lib/mixlib/install/backend/artifactory.rb
```diff
@@ -0,0 +1,70 @@
...
+            "omnibus.platform" => options.platform,
+            "omnibus.platform_version" => options.platform_version,
+            "omnibus.architecture" => options.architecture,
+            "omnibus.version" => options.product_version
+          }
+
+          headers = { "X-Result-Detail" => "properties" }
+
+          begin
+            artifact = @client.get("/api/search/prop", params, headers)
```
rhass-r7: My two cents worth:

What are you really trying to test? Are you trying to test that the endpoint itself is functioning; or the upstream artifactory gem interaction? I would argue that the case should be the latter, and as such the client.get call/response should be stubbed so that the gem interaction can be tested without external dependencies.

Having said this, there is still value in testing the endpoint. However, I am in the opinion done somewhere else, and substantially more frequently than it would be exercised here. An example of such use would be Chef Audit Mode on the boxes which are running Artifactory, as this would provide an immediate alert to an issue from a broken endpoint due to a change. 

That's my two cents though.

patrick-wright:
We will add a mock test for the unstable channel feature.

patrick-wright:
@sersut rebased from master. failed tests related to json.parse

sersut:
:+1: 

A couple of things:

1. Can we change the commit message to something that indicates the content of the change?
2. We need to support the case where we can query artifactory without the existence of p, pv & m (that would be the functionality we need to support chef-acceptance scenario) 
3. We need to also add some unit tests that check without actually talking to artifactory. 

I think we can do the merge after doing 1. 2 & 3 can come in a separate PR.

lib/mixlib/install/backend/artifactory.rb
```diff
@@ -0,0 +1,84 @@
...
+          @client = ::Artifactory::Client.new(endpoint: ARTIFACTORY_ENDPOINT)
+        end
+
+        def info
+          params = {
+            "repos" => "omnibus-current-local",
+            "omnibus.version" => options.product_version
+          }
+
+          if options.platform
```
patrick-wright: when we are only provided a platform and not the related version and arch and only returning the first result as ArtifactInfo.  I think that's ok for now, as long as there isn't a requirement to allow platform lists that include all versions and arch for unstable.  That sounds like a case I don't want to handle anyway.  Maybe the Options class needs to raise an exception if platform is provided, but pv and m are not.

sersut: I think raising an error when only `p` is provided will be better. That way we will not get usage we do not want to support moving forward :smile:

lib/mixlib/install/options.rb
```diff
@@ -21,37 +21,66 @@ class Install
...
+          errors << "Version must match pattern '1.2.3+12345678901234' when \
+using channels #{ARTIFACTORY_CHANNELS.join(", ")}"
         end
       end
 
-      SUPPORTED_OPTIONS.each do |option|
-        define_method option do
-          options[option] || options[option.to_s]
+      def validate_platform_info
+        if platform && (!platform_version || !architecture)
```
sersut: This would fail if only `platform_version` is provided. We should probably do something like:

```ruby
if (!platform.nil? || !platform_version.nil? || !architecture.nil?) && (platform.nil? || platform_version.nil? || architecture.nil?)
    raise 
end
```

patrick-wright: That was intentional.  I was trying to catch that the platform options were provided when `platform` was set, but are ignored otherwise.  If you have a strong opinion of this, I will enhance this.

sersut:
:+1: thanks @patrick-wright

sersut:
Merging since all the failures are fastly cache issues in omnitruck.

PR #18, opened 2015-10-20
Refactor Backend::Artifactory

patrick-wright:
@sersut @jtimberman 
I refactored this class to make it a bit more readable (specifically prev lines 47, 49, 50)

I also noticed the opscode s3 bucket is now https, so I fixed the tests.

This will be squashed to include two commits: the artifactory changes and the spec changes

### DIFF
--- a/lib/mixlib/install/backend/artifactory.rb
+++ b/lib/mixlib/install/backend/artifactory.rb
@@ -37,33 +37,33 @@ module Mixlib
 
         def info
           begin
-            artifact = client.get("/api/search/prop", params, headers)
+            results = client.get("/api/search/prop", params, headers)["results"]
           rescue Errno::ETIMEDOUT => e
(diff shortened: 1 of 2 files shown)


schisamo:
Is there any advantage on using [AQL](https://www.jfrog.com/confluence/display/RTF/Artifactory+Query+Language) for this search (as opposed to a property search)?

patrick-wright:
I did try the AQL post.  It was faster, but since we are making single requests with simple data it wouldn't be that noticeable.  The property search happens to return just what we need without extra data using the params we pass to the request.  The other determent was the auth setup. Not a big deal at all to set that up if we decide to go that route.  In the end, I had a simple solution and @sersut and I agreed that we could always enhance that later.  If you feel we really need to use AQL I'm open to adding a story to the backlog.

schisamo:
@patrick-wright I think the fact that a property search doesn't require authn is enough of a reason to ship this! :+1:

PR #20, opened 2015-10-20
First version of generating install command with mixlib-install.

sersut:
- [x] Change the structure of `acceptance/` to be compliant with `chef-acceptance`.
- [x] Merge https://github.com/sersut/test-kitchen/pull/1. 

This PR makes mixlib-install return an `install_command` that does not need `install.sh` to install products. This is the first part of GS-3. 

/cc: @patrick-wright


acceptance/chef/.kitchen.yml
```diff
@@ -0,0 +1,40 @@
...
+  forward_agent: yes
+  customize:
+    cpus: 2
+    memory: 1024
+
+provisioner:
+  name: chef_zero
+  product_name: chef
+  product_version: latest
+  channel: stable
```
patrick-wright: should this be `current` by default?

acceptance/chef/.kitchen.yml
```diff
@@ -0,0 +1,40 @@
...
+  #     box: chef/windows-7-professional # private
+
+suites:
+  - name: mixlib-install-chef
+    run_list:
+  - name: mixlib-install-chefdk
+    provisioner:
+      product_name: chefdk
+      product_version: latest
+      channel: stable
```
patrick-wright: Same. `current`?

acceptance/chef/.kitchen.yml
sersut: :+1: nice catch.

acceptance/chef/.kitchen.yml
sersut: ditto.

lib/mixlib/install/version.rb
```diff
@@ -1,5 +1,5 @@
 module Mixlib
   class Install
-    VERSION = "0.8.0"
+    VERSION = "1.0.0"
```
sersut: I am not sure if we should bump to 1.0.0 at this point. Since we are pretty much breaking the backwards compatibility I think we should do a major version bump. Any thoughts?

patrick-wright: I would like the change the version in master to 0.8.0-pre or 0.8.0-dev and let it be.  We can change it to whatever version we want when we release.

acceptance/Gemfile
```diff
@@ -0,0 +1,10 @@
+source "https://rubygems.org"
+
+gem "mixlib-install", path: "../"
+gem "test-kitchen", path: "../../test-kitchen"
```
schisamo: Shouldn't this come from GitHub or Rubygems.org?

lib/mixlib/install/version.rb
patrick-wright: Actually, `0.8.0.pre`.  I'm accustomed to using tools like `gem-release` https://github.com/svenfuchs/gem-release where the version increments are automated.  Since we still at a 0 major version we should keep the increments small.  I think we should stay at 0.8 while we make these deeper cuts.

lib/mixlib/install/options.rb
```diff
@@ -54,6 +54,14 @@ def validate_options!
         end
       end
 
+      def for_artifactory?
```
patrick-wright: I think there are lines in this class that could use this method

acceptance/Gemfile.lock
```diff
@@ -0,0 +1,214 @@
+GIT
```
patrick-wright: We should add the Gemfile.lcok to .gitignore since this is a gem

acceptance/current/.acceptance/acceptance-cookbook/metadata.rb
```diff
@@ -0,0 +1,7 @@
+name 'acceptance-cookbook'
+maintainer 'The Authors'
```
patrick-wright: I've been removing all but the `name` to keep these tidy :)

lib/mixlib/install/generator/bourne.rb
```diff
@@ -0,0 +1,62 @@
...
+channel=#{options.channel}
+EOS
+        end
+
+        def artifactory_urls
+          raise "not implemented yet"
+        end
+
+        def get_script(name)
+          File.read(File.join(File.dirname(__FILE__), "bourne/#{name}.sh"))
```
patrick-wright: `bourne/scripts`?

patrick-wright:
Would be nice to set up ec2 and vagrant in kitchen. Then we could run the tests in Travis (ec2) while still running them locally with vagrant.

acceptance/Gemfile.lock
sersut: I think this `Gemfile.lock` is fine because it is under `acceptance/`. This makes sure that we run the acceptance suite always with the same set of gems.

acceptance/current/.acceptance/acceptance-cookbook/metadata.rb
sersut: :+1:

lib/mixlib/install/options.rb
sersut: :+1:

acceptance/Gemfile
sersut: :+1:

lib/mixlib/install/version.rb
sersut: :+1:

sersut:
@patrick-wright probably we should leave the ec2 / vagrant config to a later PR. We might also think of enabling delivery for this and using the 'acceptance' phase there too :smile:

acceptance/Gemfile.lock
patrick-wright: Agreed. I missed that.

lib/mixlib/install/version.rb
patrick-wright: I'm wrong on this again.  Looks like the correct thing to do is leave the version untouched until it's time to release.  We should follow what chef-acceptance is doing: https://github.com/chef/chef-acceptance#releasing-chef-acceptance

sersut:
@schisamo / @patrick-wright this is now ready for the final review. 

An example usage can be seen here:

https://github.com/sersut/test-kitchen/blob/sersut/mixlib-install-update/lib/kitchen/provisioner/chef_base.rb#L278-L284

That change is part of https://github.com/sersut/test-kitchen/pull/1 which we are waiting Fletcher's return to continue merge discussions.

spec/mixlib/install/generator_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    Mixlib::Install.new(product_name: "chef",
+                        channel: channel,
+                        product_version: "latest").install_command
+  }
+
+  context "for :stable channel" do
+    let(:channel) { :stable }
+
+    it "outputs the install_command" do
+      expect(install_script).to be_a(String)
```
jtimberman: Do we want to test that the string starts with the `#!/bin/sh` and ends with the appropriate content from `install_package.sh`.

sersut: Great idea @jtimberman. Will do that.

acceptance/current/.acceptance/acceptance-cookbook/.chef/config.rb
```diff
@@ -0,0 +1 @@
+cookbook_path File.join(File.dirname(__FILE__), '..', '..')
```
schisamo: What's the purpose of this `.chef/config.rb` file?

acceptance/current/.acceptance/acceptance-cookbook/recipes/destroy.rb
```diff
@@ -0,0 +1,3 @@
+execute 'kitchen destroy' do
+  cwd File.join(File.dirname(__FILE__), "../../../../../../../..")
```
schisamo: Ugh this feels wrong. Perhaps chef-acceptance can inject some node attributes into the chef-zero run? Something like `node['chef-acceptance']['root']` or `node['chef-acceptance']['suite_root']`.

lib/mixlib/install/options.rb
```diff
@@ -71,8 +79,7 @@ def validate_channels
       end
 
       def validate_unstable_version
-        if ARTIFACTORY_CHANNELS.include?(channel) &&
-            product_version !~ /^\d+.\d+.\d+\+[0-9]{14}$/
+        if for_artifactory? && product_version !~ /^\d+.\d+.\d+\+[0-9]{14}$/
```
schisamo: So not sure we can use the version number pattern to determine the channel. Even non-integration build versions will flow through unstable. We are also moving to a place where timestamps might be dropped completely. See this RFC for more details:
https://github.com/chef/chef-rfc/blob/master/rfc047-release-process.md

acceptance/current/.acceptance/acceptance-cookbook/recipes/destroy.rb
patrick-wright: Yes.  We have a ticket to create a `kitchen_suite` resource and optimizing all this ugliness is included.

lib/mixlib/install/options.rb
patrick-wright: Ok.  We did add some version string validation.  I don't recall if we took that upon ourselves or if there was a reason for it.  Looking at it now, it's probably not necessary.

acceptance/current/.acceptance/acceptance-cookbook/.chef/config.rb
patrick-wright: without it chef doesn't recognize that the `.acceptance`dir the cookbooks dir.  Any advice were be appreciated.

lib/mixlib/install/options.rb
sersut: We added this because we were running into issues where we get multiple candidates from artifactory when you specify a version like `12.5.1` only. I recommend we merge with this logic and do the necessary when we are doing the minor changes when we enable :unstable channel in the artifactory for reals.

acceptance/current/.acceptance/acceptance-cookbook/recipes/destroy.rb
sersut: +1 for what @patrick-wright is saying. Once we have a few projects using chef-acceptance and we know what they need, we will introduce shortcuts.

acceptance/current/.acceptance/acceptance-cookbook/.chef/config.rb
schisamo: Yeah the [chef-acceptance embedded `chef-client` run](https://github.com/chef/chef-acceptance/blob/master/lib/chef-acceptance/test_suite.rb#L45-L73) should render a temporary `config.rb` on the fly like `delivery-cli` does:
https://github.com/chef/delivery-cli/blob/7acaafed2291e644eed97fdda07ac634ad5f8d6f/src/delivery/job/workspace.rs#L51-L68

acceptance/current/.acceptance/acceptance-cookbook/recipes/destroy.rb
schisamo: I think it's larger than just the kitchen execution (and any associated chef resources). The `chef-client` run that the chef-acceptance CLI command (or eventual resource) executes should have a `dns.json` with useful node attribute data. This is the pattern that delivery-cli follows with great success:
https://github.com/chef/delivery-cli/blob/7acaafed2291e644eed97fdda07ac634ad5f8d6f/src/delivery/job/workspace.rs#L457-L480

sersut: Definitely agree with the comment @schisamo. We will address that in chef-acceptance first and make the necessary changes in mixlib-install once it is done. I have captured that feedback in GS-16.

acceptance/current/.acceptance/acceptance-cookbook/.chef/config.rb
sersut: Agreed. We will take care of that under the context of GS-16.

Issue #16, opened 2015-10-14
invalid version detection in install_command.ps1 


hh:
As a chef-provisioning user
When I set chef_version in convergence_options to 12.5.1
Then I expect a node with 12.4.1 installed to upgraded to 12.5.1.

(If this needs to be a chef-provisioning ticket, let me know)

Suspecting something around https://github.com/chef/mixlib-install/blob/14611366684808e4db54dfaeee8ca49a8816a4e6/support/install_command.ps1#L12

```
               convergence_options: {
                 chef_version: '12.5.1'
               },
```

```
    - write file C:\chef\client.rb on base-2012-hardened-xviii
    - write file C:\chef\install.ps1 on base-2012-hardened-xviii
    [base-2012-hardened-xviii] Method invocation failed because [Selected.System.String] does not contain a 
                               method named 'StartsWith'.
    [base-2012-hardened-xviii] -----> Chef Omnibus installation detected (12.5.1)
                               
    [base-2012-hardened-xviii] At C:\chef\install.ps1:19 char:7
                               +   if ($chef_version.StartsWith($version)) { return $false }
                               +       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                   + CategoryInfo          : InvalidOperation: (StartsWith:String) [], Runtim 
                                  eException
                                   + FullyQualifiedErrorId : MethodNotFound
                                
    - run 'powershell.exe -ExecutionPolicy Unrestricted -NoProfile "& ""C:\chef\install.ps1"""' on base-2012-hardened-xviii[2015-10-14T09:46:01-04:00] INFO: Processing chef_node[base-2012-hardened-xviii] action create (basic_chef_client::block line 57)
[2015-10-14T09:46:02-04:00] INFO: Converging base-2012-hardened-xviii because the resource was updated ...

    [base-2012-hardened-xviii] [2015-10-14T13:46:18+00:00] INFO: *** Chef 12.4.1 ***
```

thommay:
@jaym @smurawski can y'all have a look at this? I lifted the powershell code from kitchen I think.

jaym:
maybe related to https://github.com/test-kitchen/test-kitchen/pull/864

dissonanz:
Yes it's the same. I had to refactor the split in order to stay within the command size limit, but this seems to work.

schisamo:
Fix is in https://github.com/chef/mixlib-install/pull/21

patrick-wright:
[...]

PR #21, opened 2015-10-26
Fix version detection in windows #16

dissonanz:


schisamo:
Looks reasonable to me. Can you give things a review @patrick-wright @sersut @jaym?

jaym:
:+1:

patrick-wright:
@thommay Can you please review this change before we merge?

thommay:
:+1:

mwrock:
:-1:

mwrock:
:+1:

sersut:
Thanks @dissonanz. :+1: 

One thing to keep in mind is that we have moved the logic that uses these files under `Mixlib::Install::ScriptGenerator` since `0.8.0.alpha.0` release. I wonder if this is a blocker for many @thommay. In that case we should probably release a `0.7.1`. Otherwise we will get this out with the next alpha release of `0.8.0` which will eventually become `1.0.0` when we finish our rehaul. 

Let us know your thoughts on releasing this @thommay / @dissonanz.

thommay:
hm, we should probably release this as a `0.7.1` I suppose - but I don't know how much this is affecting people.

cheeseplus:
+1

patrick-wright:
This patch was applied to a branch from the 0.7.0 release.  0.7.1 is tagged, and has been published to rubygems.org

PR #22, opened 2015-11-11
fixes windows metadata url when version is true

mwrock:
Today when running test-kitchen off master, windows runs break with the default chef version because "true" is given as the version param to the metadata url.

### DIFF
--- a/lib/mixlib/install/script_generator.rb
+++ b/lib/mixlib/install/script_generator.rb
@@ -107,7 +107,8 @@ module Mixlib
       # @return [String] shell variable lines
       # @api private
       def install_command_vars_for_bourne
-        flags = %w[latest true nightlies].include?(version) ? "" : "-v #{CGI.escape(version)}"
+        flags = ""
+        flags << "-v #{version_metadata_flag}" if version_metadata_flag
         flags << " " << "-n" if nightlies
         flags << " " << "-p" if prerelease
         flags << " " << install_flags if install_flags
@@ -120,6 +121,10 @@ module Mixlib
           shell_var("sudo_sh", sudo("sh")),
           shell_var("version", version)
         ].join("\n")
+      end
+
+      def version_metadata_flag
+        %w[latest true nightlies].include?(version) ? nil : CGI.escape(version)
       end
 
       # Generates the install command variables for PowerShell-based platforms.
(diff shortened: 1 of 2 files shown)


mwrock:
To add more clarity here: this PR simply applies the same logic to the `-v` parameter of the metadate url for Windows as it has been doing for bourne shells.

schisamo:
@mwrock Thanks for the clarification. I would like @patrick-wright and @sersut to take a look at this as they have been performing heavy modifications in this code base recently.

patrick-wright:
Unfortunately this won't fix the issue given the project's current state.  test-kitchen depends on version 0.7, and the `Mixlib::Install` constructor has changed for `mixlib-install:master`.   What used to be the `Install` class is now the `ScriptGenerator`class which will actually be removed in the very near future.  I suppose we could branch from 0.7 and create an 0.7.1 patch release, but @sersut and I are actively working on releasing the next version.  See PR https://github.com/chef/mixlib-install/pull/20 for more details.

mwrock:
You are correct, but 0.8 is not released on rubygems. I have a hack in test kitchen until this is released. After this is released I'll switch it over to use ScriptGenerator.

mwrock:
@patrick-wright would I be better off submitting this to your unstable-install-command branch?

patrick-wright:
@mwrock, one of the issues with ScriptGenerator is that it builds the omnitruck url using the old params.  Serdar and I are back next week, and we'll discuss this PR on Monday.

@schisamo Can you provide some context to the previous omnitruck params?  Are those going to be backward compatible moving forward or will those be deprecated at some point?

mwrock:
My apologies. This is what happens when responding to PR commnts and brushing teeth at the same time. I missed the part about removing `ScriptGenerator`. Once your branch is imminent, it would be great if you could ping me and I'd be happy to test it and also submit a PR to test-kitchen to retrofit to using the updated gem.

patrick-wright:
@mwrock test-kitchen is a big driver of this work.  Check this out: https://github.com/sersut/test-kitchen/pull/1

mwrock:
ok. got it. And I have https://github.com/test-kitchen/test-kitchen/pull/878 which addresses issues whith the generated powershell script exceeding the winrm commandline limit.

patrick-wright:
The metadata flags are no longer used this way with the introduction of omnitruck channels.

PR #25, opened 2015-12-02
Move product matrix and its documentation from chef-ingredient.

sersut:
This PR moves [product matrix](https://github.com/chef-cookbooks/chef-ingredient/blob/master/PRODUCT_MATRIX.md) and its helper methods from chef-cookbooks/chef-ingredient to here. A few things to keep in mind while reviewing:

1. I do not think the `package_name` column in the `PRODUCT_MATRIX.md` makes sense. I think we should add a description column for it to be more useful for the end user. But I did not do that and copied over only what we have. 
2. @jtimberman can you double check the version based logic in the matrix. I have double checked it but just in case :smile: 

/cc: @chef/engineering-services, @thommay

### DIFF
--- a/PRODUCT_MATRIX.md
+++ b/PRODUCT_MATRIX.md
@@ -1,20 +1,20 @@
-| Product | Product Key  | Package Name |
-| ------- | ------------ | ------------ |
-| Chef Client | chef | chef |
-| Chef Development Kit | chefdk | chefdk |
-| Chef Server | chef-server | chef-server-core |
-| Management Console | manage | chef-manage |
-| Chef Server High Availability addon | chef-ha | chef-ha |
-| Chef Server Reporting addon | reporting | opscode-reporting |
-| Chef Cloud Marketplace addon | supermarket | supermarket |
-| Chef Cloud Marketplace addon | chef-marketplace | chef-marketplace |
-| Chef Server Replication addon | chef-sync | chef-sync |
-| Delivery | delivery | delivery |
-| Delivery CLI | delivery-cli | delivery-cli |
-| Analytics Platform | analytics | opscode-analytics |
-| Chef Compliance | compliance | chef-compliance |
-| Chef Push Server | push-server | push-jobs-server |
-| Chef Push Server | push-client | push-jobs-client |
-| Enterprise Chef (legacy) | private-chef | private-chef |
+| Product | Product Key  |
+| ------- | ------------ |
+| Analytics Platform | analytics |
+| Chef Client | chef |
+| Chef Server High Availability addon | chef-ha |
+| Chef Cloud Marketplace addon | chef-marketplace |
+| Chef Server | chef-server |
+| Chef Server Replication addon | chef-sync |
+| Chef Development Kit | chefdk |
+| Chef Compliance | compliance |
+| Delivery | delivery |
+| Delivery CLI | delivery-cli |
+| Management Console | manage |
+| Enterprise Chef (legacy) | private-chef |
+| Chef Push Server | push-client |
+| Chef Push Server | push-server |
+| Chef Server Reporting addon | reporting |
+| Supermarket | supermarket |
 
 Do not modify this file manually. It is automatically rendered via a rake task.
--- a/PRODUCT_MATRIX.md
+++ b/PRODUCT_MATRIX.md
@@ -1,0 +1,20 @@
+| Product | Product Key  | Package Name |
+| ------- | ------------ | ------------ |
+| Chef Client | chef | chef |
+| Chef Development Kit | chefdk | chefdk |
+| Chef Server | chef-server | chef-server-core |
+| Management Console | manage | chef-manage |
+| Chef Server High Availability addon | chef-ha | chef-ha |
+| Chef Server Reporting addon | reporting | opscode-reporting |
+| Chef Cloud Marketplace addon | supermarket | supermarket |
+| Chef Cloud Marketplace addon | chef-marketplace | chef-marketplace |
+| Chef Server Replication addon | chef-sync | chef-sync |
+| Delivery | delivery | delivery |
+| Delivery CLI | delivery-cli | delivery-cli |
+| Analytics Platform | analytics | opscode-analytics |
+| Chef Compliance | compliance | chef-compliance |
+| Chef Push Server | push-server | push-jobs-server |
+| Chef Push Server | push-client | push-jobs-client |
+| Enterprise Chef (legacy) | private-chef | private-chef |
+
+Do not modify this file manually. It is automatically rendered via a rake task.
--- a/Rakefile
+++ b/Rakefile
@@ -17,3 +17,22 @@ end
 
 desc "Run all tests"
 task test: [:rubocop, :spec]
+
+desc "Render product matrix documentation"
+task "matrix" do
+  require "mixlib/install/product"
+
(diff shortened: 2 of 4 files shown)


curiositycasualty:
Seems like the travis failure is actually https://github.com/chef/mixlib-versioning 's fault.

:+1:

PRODUCT_MATRIX.md
```diff
@@ -0,0 +1,20 @@
...
+| Chef Server Replication addon | chef-sync | chef-sync |
+| Delivery | delivery | delivery |
+| Delivery CLI | delivery-cli | delivery-cli |
+| Analytics Platform | analytics | opscode-analytics |
+| Chef Compliance | compliance | chef-compliance |
+| Chef Push Server | push-server | push-jobs-server |
+| Chef Push Server | push-client | push-jobs-client |
+| Enterprise Chef (legacy) | private-chef | private-chef |
+
+Do not modify this file manually. It is automatically rendered via a rake task.
```
jtimberman: :rocket: :cake:

PRODUCT_MATRIX.md
```diff
@@ -0,0 +1,20 @@
+| Product | Product Key  | Package Name |
+| ------- | ------------ | ------------ |
+| Chef Client | chef | chef |
```
patrick-wright: I agree.  Package Name is really internal to the resource, and should be removed from the md.

lib/mixlib/install/product.rb
```diff
@@ -0,0 +1,259 @@
...
+    end
+    config_file "/etc/opscode-push-jobs-server/opscode-push-jobs-server.rb"
+  end
+
+  product "push-client" do
+    product_name "Chef Push Server"
+    package_name do |v|
+      v < version_for("1.3.0") ? "opscode-push-jobs-client" : "push-jobs-client"
+    end
+  end
```
schisamo: This should be alphabetized by product key

lib/mixlib/install/product.rb
```diff
@@ -0,0 +1,259 @@
...
+require "mixlib/versioning"
+
+module Mixlib
+  class Install
+    class Product
+      def initialize(&block)
+        instance_eval(&block)
+      end
+
+      DSL_PROPERTIES = [:product_name, :package_name, :ctl_command, :config_file]
```
schisamo: alphabetized with a property per line:

```ruby
DSL_PROPERTIES = [
  :config_file,
  :ctl_command,
  :package_name,
  :product_name, 
]
```

Rakefile
```diff
@@ -17,3 +17,22 @@ end
...
+task "matrix" do
+  require "mixlib/install/product"
+
+  doc_file = File.join(File.dirname(__FILE__), "PRODUCT_MATRIX.md")
+  puts "Updating doc file at: #{doc_file}"
+
+  File.open(doc_file, "w+") do |f|
+    f.puts("| Product | Product Key  | Package Name |")
+    f.puts("| ------- | ------------ | ------------ |")
+    PRODUCT_MATRIX.products.each do |p_key|
```
schisamo: Sort by product name before calling each.

sersut:
@chef/engineering-services ordering, travis issues are fixed now. Also removed `package name` from documentation.

lib/mixlib/install/product.rb
```diff
@@ -0,0 +1,264 @@
...
+
+  product "reporting" do
+    product_name "Chef Server Reporting addon"
+    package_name "opscode-reporting"
+    ctl_command "opscode-reporting-ctl"
+    config_file "/etc/opscode-reporting/opscode-reporting.rb"
+  end
+
+  product "supermarket" do
+    product_name "Chef Cloud Marketplace addon"
```
schisamo: copy pasta error

schisamo:
@sersut one more comment and the products were not sorted per https://github.com/chef/mixlib-install/pull/25#discussion-diff-46471585

PRODUCT_MATRIX.md
```diff
@@ -0,0 +1,20 @@
...
+| Chef Development Kit | chefdk |
+| Chef Compliance | compliance |
+| Delivery | delivery |
+| Delivery CLI | delivery-cli |
+| Management Console | manage |
+| Enterprise Chef (legacy) | private-chef |
+| Chef Push Server | push-client |
+| Chef Push Server | push-server |
+| Chef Server Reporting addon | reporting |
+| Chef Cloud Marketplace addon | supermarket |
```
jtimberman: > | Supermarket | supermarket |

:grin:

lib/mixlib/install/product.rb
```diff
@@ -0,0 +1,264 @@
...
+              if value.is_a? String
+                value
+              else
+                value.call(version_for(version))
+              end
+            else
+              instance_variable_set("@#{prop}".to_sym, prop_string)
+            end
+          else
+            raise "Not both at the same time." if !prop_string.nil?
```
jtimberman: This seems ambiguous. Both what?

sersut: :+1: Fixed now :smile:

PRODUCT_MATRIX.md
sersut: Nice catch. @schisamo caught it too. Fixed now.

sersut:
@schisamo thanks for the through look :smile: Fixed now.

lib/mixlib/install/product.rb
```diff
@@ -0,0 +1,264 @@
...
+  end
+
+  product "chef-server" do
+    product_name "Chef Server"
+    package_name do |v|
+      if (v < version_for("12.0.0")) && (v > version_for("11.0.0"))
+        "chef-server"
+      else
+        "chef-server-core"
+      end
```
jtimberman: Don't we need something to account for this, from chef-ingredient's helpers:

```ruby
elsif (v < Mixlib::Versioning.parse('11.0.0')) && (v > Mixlib::Versioning.parse('1.0.0'))
 Chef::Log.fatal("Invalid version specified, '#{version}' for #{product}!")
  fail
end
```

sersut: @jtimberman what is the purpose of that? is it so that we do not get anything `< 11.0.0`?

spec/mixlib/install/product_spec.rb
```diff
@@ -0,0 +1,254 @@
...
+    end
+
+    it "returns nil for unset properties" do
+      expect(product.package_name).to eq(nil)
+    end
+
+    it "errors for unexisting properties" do
+      expect { product.address }.to raise_error(StandardError)
+    end
+  end
```
jtimberman: > TODO: create a custom error handler for invalid properties

Perhaps?

lib/mixlib/install/product.rb
jtimberman: Comment from that conditional block in the chef-ingredient code:

```
      # We want to validate that we're getting a version that is valid
      # for the Chef Server. However, since the default is 0.0.0,
      # implying latest, we need to additionally ensure that the
      # bottom version is something valid. If we don't have the check
      # in the elsif, it will say that 0.0.0 is not a valid version.
```

spec/mixlib/install/product_spec.rb
sersut: Meh... Probably that would be a TODO that lives for a long time so I say let's not add it.  :smile:

PR #27, opened 2015-12-04
Windows Support

patrick-wright:
first pass for supporting windows install scripts for omnitruck.

removed nightlies and prerelease switches

artifactory urls for unstable channel not yet implemented

todo:
- ~~optimize powershell install_command~~
- ~~refactor paths for Install #root and #current_version methods (and make sure they are correct)~~
- ~~validate with chef-acceptance (kitchen)~~
- ~~implement in chef-ingredient~~

### DIFF
--- a/acceptance/current/.kitchen.yml
+++ b/acceptance/current/.kitchen.yml
@@ -23,12 +23,9 @@ platforms:
   # - name: macosx-10.10
   #   driver:
   #     box: chef/macosx-10.10 # private
-  # - name: windows-7-professional
-  #   provisioner:
-  #     name: windows_chef_zero
-  #     require_chef_omnibus: 11.12.4
-  #   driver:
-  #     box: chef/windows-7-professional # private
+  - name: windows-server-2012r2-standard
+    driver:
+      box: chef/windows-server-2012r2-standard # private
 
 suites:
   - name: mixlib-install-chef
--- a/lib/mixlib/install.rb
+++ b/lib/mixlib/install.rb
@@ -56,10 +56,13 @@ module Mixlib
     # @return [String] the installation directory for the project
     #
     def root
-      # TODO: Support root as "$env:systemdrive\\opscode\\chef" when on windows.
       # This only works for chef and chefdk but they are the only projects
       # we are supporting as of now.
-      "/opt/#{options.product_name}"
+      if options.for_ps1?
+        "$env:systemdrive\\opscode\\#{options.product_name}"
+      else
+        "/opt/#{options.product_name}"
+      end
     end
 
     #
@@ -72,7 +75,12 @@ module Mixlib
       # install directory which can be different than the product name (e.g.
       # chef-server -> /opt/opscode). But this is OK for now since
       # chef & chefdk are the only supported products.
-      version_manifest_file = "/opt/#{options.product_name}/version-manifest.json"
+      version_manifest_file = if options.for_ps1?
+        "$env:systemdrive\\opscode\\#{options.product_name}\\version-manifest.json"
+      else
+        "/opt/#{options.product_name}/version-manifest.json"
+      end
+
       if File.exist? version_manifest_file
         JSON.parse(File.read(version_manifest_file))["build_version"]
       end
--- a/lib/mixlib/install/generator/powershell.rb
+++ b/lib/mixlib/install/generator/powershell.rb
@@ -1,0 +1,67 @@
+#
+# Copyright:: Copyright (c) 2015 Chef, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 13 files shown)


patrick-wright:
@chef/engineering-services

acceptance/current/.kitchen.yml
```diff
@@ -23,12 +23,12 @@ platforms:
   # - name: macosx-10.10
   #   driver:
   #     box: chef/macosx-10.10 # private
-  # - name: windows-7-professional
-  #   provisioner:
-  #     name: windows_chef_zero
-  #     require_chef_omnibus: 11.12.4
-  #   driver:
-  #     box: chef/windows-7-professional # private
+  - name: windows-7-professional
```
schisamo: After chatting with some of our Windows team they recommended standardizing on Windows Server 2012R2 for our default Windows test platform. Take a look at:
https://github.com/chef-cookbooks/omnibus/blob/master/.kitchen.yml#L57-L59

lib/mixlib/install/generator/powershell.rb
```diff
@@ -0,0 +1,58 @@
...
+          if options.for_artifactory?
+            raise "not implemented yet"
+            # install_command << get_script(:get_project_metadata_for_artifactory)
+          else
+            install_command << get_script(:get_project_metadata)
+          end
+          install_command << get_script(:install_project)
+          install_command << "}"
+          install_command << render_command
+          install_command.join("\n\n")
```
schisamo: Would it be easier to use ERB actual templates for all of this?

sersut: We would like to keep different parts of the script in different files @schisamo so that in the future we can add methods that can render only the part of the script. I think we can do a little bit of a better job here organizing things but I am against conversion to templates since we would like to keep things separate.

sersut: I would do something like this @patrick-wright:

```
install_project_module = []
install_project_module << get_script(:helpers)
if options.for_artifactory?
  raise "not implemented yet"
  # install_project_module << get_script(:get_project_metadata_for_artifactory)
else
  install_project_module << get_script(:get_project_metadata)
end
install_project_module << get_script(:install_project)

install_command = []
install_command << ps1_modularize(install_project_module.join("\n\n"), "Omnitruck")
install_command << render_command
install_command.join("\n\n")

def ps1_modularize(module_body, module_name)
  ps1_module = []
  ps1_module << "new-module -name #{module_name} -scriptblock {"
  ps1_module << module_body
  ps1_module << "}"
  ps1_module.join("\n\n")
end
```

lib/mixlib/install.rb
```diff
@@ -56,10 +56,11 @@ def install_command
     # @return [String] the installation directory for the project
     #
     def root
-      # TODO: Support root as "$env:systemdrive\\opscode\\chef" when on windows.
```
sersut: We can remove this TODO line.

lib/mixlib/install/generator/powershell.rb
```diff
@@ -0,0 +1,67 @@
...
+
+        def ps1_modularize(module_body, module_name)
+          ps1_module = []
+          ps1_module << "new-module -name #{module_name} -scriptblock {"
+          ps1_module << module_body
+          ps1_module << "}"
+          ps1_module.join("\n")
+        end
+
+        def render_command
```
sersut: I would rename this to `install_command` and have it as a single line instead of using block string comment.

lib/mixlib/install/options.rb
```diff
@@ -28,11 +28,20 @@ class InvalidOptions < ArgumentError; end
...
+        :platform,
+        :platform_version,
+        :product_name,
+        :product_version,
+        :shell_type
+      ]
 
       def initialize(options)
         @options = options
+        @shell_type = options.fetch(:shell_type, :sh)
```
sersut: I don't think we need to store the shell_type in an instance_variable here.

lib/mixlib/install/options.rb
```diff
@@ -28,11 +28,20 @@ class InvalidOptions < ArgumentError; end
...
-                           :platform, :platform_version, :architecture]
+      SUPPORTED_SHELL_TYPES = [:ps1, :sh]
+      SUPPORTED_OPTIONS = [
+        :architecture,
+        :channel,
+        :platform,
+        :platform_version,
+        :product_name,
+        :product_version,
+        :shell_type
```
sersut: Let's add a basic spec for `shell_type` to `options_spec` to catch possible regressions in the future.

sersut:
Super sweet @patrick-wright. Some minor comments. 

Also can we squash with a good commit message. 

Other than those :shipit:

lib/mixlib/install/generator/powershell.rb
patrick-wright: we already have `install_command`.  How about `exec_command` or `run_command`?

patrick-wright: or did you mean move it into install_command?

patrick-wright: @sersut let's just  keep it this way.  Then I'll have to deal with rubocop errors.

sersut:
:+1: 

Only one `TODO` line to remove. Also can we merge the `test-kitchen` PR first and fix up the Gemfile & Gemfile.lock here before the merge.

PR #28, opened 2015-12-04
Remove version checks for push server 2.0

jtimberman:
At this time we don't have a Push Server 2.0 - the "latest" is 1.1.6.
The logic for :latest setting the version to 1000.1000.1000 means
we'll always be > 2.0.


sersut:
:+1:

sersut:
Err... Actually I just remembered I saw this: https://github.com/chef-cookbooks/chef-ingredient/pull/72/files

schisamo:
:shipit:

sersut:
Checking `artifactory.chef.co` seems like this PR got it right.

PR #35, opened 2015-12-14
support latest version keyword for unstable channel

patrick-wright:
Adds support for using `:latest` (code), or `'latest'` (kitchen.yml) when using the unstable channel against artifactory.

@jtimberman @sersut @schisamo

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,9 +35,16 @@ artifacts.first.url
 # => "http://opscode-omnibus-packages-current.s3.amazonaws.com/mac_os_x/10.9/x86_64/chef-12.5.1%2B20151009083009-1.dmg"
 ```
 
+## Unstable channel
+The `:unstable` channel is currently only available when connected to Chef's internal network.
+Configure Artifactory access by setting the following environment variables:
+```
+export ARTIFACTORY_USERNAME='username'
+export ARTIFACTORY_PASSWORD='password'
+```
 
-## Test
-Some tests are tagged `:unstable` and can only run when connected to Chef's internal network.  These are excluded by default.  To run the `:unstable` tests run: `bundle exec rspec --tag unstable`.
+### Unstable channel specs
+Some spec examples are tagged `:unstable` and can only run when connected to Chef's internal network.  These are excluded by default.  To run the `:unstable` tests run: `bundle exec rspec --tag unstable`.
 
 ## Contributing
 
--- a/acceptance/unstable/.kitchen.yml
+++ b/acceptance/unstable/.kitchen.yml
@@ -28,13 +28,13 @@ suites:
   - name: chef-unstable-install
     provisioner:
       product_name: chef
-      product_version: 12.5.1+20151210002019
+      product_version: latest
       channel: unstable
     run_list:
   - name: chefdk-unstable-install
     excludes: [ 'freebsd-10.0' ]
     provisioner:
       product_name: chefdk
-      product_version: 0.10.0+20151209030614
+      product_version: latest
       channel: unstable
     run_list:
--- a/lib/mixlib/install/backend/artifactory.rb
+++ b/lib/mixlib/install/backend/artifactory.rb
@@ -25,66 +25,98 @@ module Mixlib
   class Install
     class Backend
       class Artifactory
-        ARTIFACTORY_ENDPOINT = "http://artifactory.chef.co".freeze
-
-        ARTIFACTORY_REPOSITORY = "omnibus-unstable-local".freeze
-
-        attr_reader :options
-        attr_reader :client
+        attr_accessor :options
 
         def initialize(options)
           @options = options
-          @client = ::Artifactory::Client.new(endpoint: ARTIFACTORY_ENDPOINT)
         end
 
+        # Create filtered list of artifacts
+        #
+        # @return [Array<ArtifactInfo>] list of artifacts for the configured
+        # channel, product name, and product version.
+        # @return [ArtifactInfo] arifact info for the configured
+        # channel, product name, product version and platform info
+        #
         def info
-          begin
-            results = client.get("/api/search/prop", params, headers)["results"]
-          rescue Errno::ETIMEDOUT => e
-            raise e, "unstable channel uses endpoint #{ARTIFACTORY_ENDPOINT} \
-which is currently only accessible through Chef's internal network."
+          artifacts = artifactory_info.collect { |a| create_artifact(a) }
+          artifacts_for_version = artifacts.find_all do |a|
+            a.version == options.resolved_version(artifacts)
           end
 
           if options.platform
-            artifact(results.first)
-          else
-            results.collect do |result|
-              artifact(result)
+            artifacts_for_version.find do |a|
+              a.platform == options.platform &&
+                a.platform_version == options.platform_version &&
+                a.architecture == options.architecture
             end
+          else
+            artifacts_for_version
           end
+        end
+
+        # Fetches all artifacts from the configured Artifactory repository using
+        # channel and product name as search criteria
+        #
+        # @return [Array<Hash>] list of artifactory hash data
+        #
+        # Hash data:
+        #   download_uri: The full url download path
+        #   <property_name>: The names of the properties associcated to the artifact
+        #
+        def artifactory_info
+          query = <<-QUERY
+items.find(
+  {"repo": "omnibus-#{options.channel}-local"},
+  {"@omnibus.project": "#{options.product_name}"}
+).include("repo", "path", "name", "property")
+          QUERY
+
+          # Creates a client based on ENVs:
+          #   ARTIFACTORY_ENDPOINT
+          #   ARTIFACTORY_USERNAME
+          #   ARTIFACTORY_PASSWORD
+          client = ::Artifactory::Client.new(
+            endpoint: ENV["ARTIFACTORY_ENDPOINT"] || "http://artifactory.chef.co"
+          )
+
+          results = client.post("/api/search/aql",
+                                query, "Content-Type" => "text/plain")
+
+          results["results"].collect do |result|
+            # Construct the downloadUri from artifact metadata
+            uri = []
+            uri << ::Artifactory.endpoint
+            uri << result["repo"]
+            uri << result["path"]
+            uri << result["name"]
+            # Merge artifactory properties and downloadUri to a flat Hash
+            { "downloadUri" => uri.join("/") }.merge(map_properties(result["properties"]))
+          end
+        end
+
+        def create_artifact(artifact_map)
+          ArtifactInfo.new(
+            md5:              artifact_map["omnibus.md5"],
+            sha256:           artifact_map["omnibus.sha256"],
+            version:          artifact_map["omnibus.version"],
+            platform:         artifact_map["omnibus.platform"],
+            platform_version: artifact_map["omnibus.platform_version"],
+            architecture:     artifact_map["omnibus.architecture"],
+            url:              artifact_map["downloadUri"]
+          )
         end
 
         private
 
-        def artifact(result)
-          ArtifactInfo.new(
-            md5:              result["properties"]["omnibus.md5"].first,
-            sha256:           result["properties"]["omnibus.sha256"].first,
-            version:          result["properties"]["omnibus.version"].first,
-            platform:         result["properties"]["omnibus.platform"].first,
-            platform_version: result["properties"]["omnibus.platform_version"].first,
-            architecture:     result["properties"]["omnibus.architecture"].first,
-            url:              result["downloadUri"]
-          )
-        end
-
-        def params
-          params = {
-            "repos" => ARTIFACTORY_REPOSITORY,
-            "omnibus.version" => options.product_version
-          }
-
-          if options.platform
-            params["omnibus.platform"] = options.platform
-            params["omnibus.platform_version"] = options.platform_version
-            params["omnibus.architecture"] = options.architecture
+        # Converts Array<Hash> where the Hash is a key pair and
+        # value pair to a simplifed key/pair Hash
+        #
+        def map_properties(properties)
+          return {} if properties.nil?
+          properties.each_with_object({}) do |prop, h|
+            h[prop["key"]] = prop["value"]
           end
-
-          params
-        end
-
-        def headers
-          { "X-Result-Detail" => "info, properties" }
         end
       end
     end
--- a/lib/mixlib/install/generator/powershell.rb
+++ b/lib/mixlib/install/generator/powershell.rb
@@ -65,14 +65,26 @@ module Mixlib
         end
 
         def artifactory_urls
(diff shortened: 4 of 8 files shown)


lib/mixlib/install/backend/artifactory.rb
```diff
@@ -54,24 +56,31 @@ def info
...
         end
 
-        private
+        def product_versions
+          params = {
+            g: "com.getchef",
+            a: options.product_name,
+            repos: ARTIFACTORY_REPOSITORY
+          }
+          results = client.get("/api/search/versions", params)["results"]
```
schisamo: I'm really worried about using any GAVC search logic in Artifactory. I would prefer we return a list of all known versions for a project (in a particular channel) using AQL (query with `omnibus.project` and return only unique `omnibus.version`). After you have that list it's possible to parse/sort everything with mixlib-versioning.

We use a similar query in chef-web-downloads:
https://github.com/chef/chef-web-downloads/blob/master/lib/chef/web/downloads/package_list.rb#L35-L43

lib/mixlib/install/backend/artifactory.rb
```diff
@@ -54,24 +56,31 @@ def info
...
-            platform_version: result["properties"]["omnibus.platform_version"].first,
-            architecture:     result["properties"]["omnibus.architecture"].first,
-            url:              result["downloadUri"]
-          )
+          results.collect { |result| result["version"] }
+        end
+
+        def product_version
+          if options.latest_version?
+            Mixlib::Versioning.find_target_version(product_versions, nil, false, true).to_s
```
sersut: This is sweet @patrick-wright!

In omnitruck we determine the version like this https://github.com/chef/omnitruck/blob/master/lib/chef/version_resolver.rb#L170-L174 . We might want to look into mixlib-versioning and see if there is anything different. Maybe we can provide the same support as omnitruck does. 

Awesome job!

patrick-wright: The Opcode::Version `find_target_version` method includes prerelease and nightlies logic.  I don't think we need that.  I guess I'm not sure what the current line is missing.  What is it you are seeing as an improvement?

lib/mixlib/install/backend/artifactory.rb
patrick-wright: In this particular case using group id and artifact id will likely give us some mileage.  I chose this because it's an anonymous request.  However, now that we are introducing additional Artifactory queries it's time to implement the AQL post and along with authentication.  It would also improve error handing up front when using the unstable channel.

Are you okay with allowing this to merge, and I'll start a new PR immediately to refactor all AQL stuff?

patrick-wright:
tests are failing due to tests that need to be properly tagged

lib/mixlib/install/backend/artifactory.rb
schisamo: Sure. I just don't want us wasting time on why GAVC logic is not sorting correctly. FWIW the Artifactory team is mostly ignoring GAVC based queries and encouraging heavy use of AQL and property based searches. Pretty sure the newer versions of Artifactory are being optimize for this approach.

schisamo:
So a general thought/idea for you @patrick-wright. We need to begin determining the public interface for the backend classes. I would have the backends extend from `Mixlib::Install::Backend` and then implement a series of abstract instance methods like this:

```ruby
#
# @abstract
#
def info
  raise NotImplementedError
end
```

This is the approach we took with Omnibus's publishers and packagers and has worked out well.

It's also time to begin introducing some YARD documentation for all the classes and methods. This is very important in a shared repository like `mixlib-install`.

lib/mixlib/install/backend/artifactory.rb
patrick-wright: Ok cool. Note that I am not using Artifactory's latest version fetching. It's fetching all versions for a channel/product.

patrick-wright:
@schisamo @sersut Okay here's another version.  It really did make sense to make the AQL changes in this PR.

First thing, it requires that the `ARTIFACTORY_*` env vars are set. Docs in progress.

Acceptance tests pass, but I really need to do more validation to make sure the versions I expect are being installed.  The specs are good, but this was a big change.

It could use improved code docs and probably some readability love.

acceptance/unstable/.kitchen.yml
```diff
@@ -38,3 +38,16 @@ suites:
       product_version: 0.10.0+20151209030614
       channel: unstable
     run_list:
+  - name: chef-unstable-install-latest
+    provisioner:
+      product_name: chef
+      product_version: latest
```
sersut: Rather than adding a new suite (which multiplies the number of test cases) let's switch chefdk to use `:latest`.

lib/mixlib/install/backend/artifactory.rb
```diff
@@ -19,72 +19,141 @@
...
+        #
+        # into
+        #
+        #  {
+        #    "some_key_name" => "some_value",
+        #    "some_key_name" => "some_value"
+        #  }
+        #
+        def map_properties(properties)
+          properties.each_with_object({}) do |prop, h|
```
sersut: Let's protect against `properties` being `nil` and return nil.

lib/mixlib/install/backend/artifactory.rb
```diff
@@ -19,72 +19,141 @@
...
+        end
 
-          if options.platform
-            artifact(results.first)
+        #
+        # Parse latest version
+        #
+        def product_version
+          if options.latest_version?
+            Mixlib::Versioning.find_target_version(product_versions, nil, false, true).to_s
```
sersut: We should have some documentation here about the parameters we are passing in to `find_target_version`.

lib/mixlib/install/backend/artifactory.rb
```diff
@@ -19,72 +19,141 @@
...
     class Backend
-      class Artifactory
-        ARTIFACTORY_ENDPOINT = "http://artifactory.chef.co".freeze
+      class Artifactory < Base
+        #
+        # Returns Array<ArtifactInfo> when platform info is not configured
+        # Returns ArtifactInfo when platform info is configured
+        #
+        def info
+          if options.platform
```
sersut: Definitely needs some readability and improvements. Looks like we are calling `products` as well as `product_version` regardless of the case. With the current separation of logic into functions it is very hard to follow.

lib/mixlib/install/backend/artifactory.rb
```diff
@@ -19,72 +19,141 @@
...
         end
 
-        def params
-          params = {
-            "repos" => ARTIFACTORY_REPOSITORY,
-            "omnibus.version" => options.product_version
-          }
+        #
+        # Creates a client based on the required ENVs:
+        #   ARTIFACTORY_ENDPOINT
```
sersut: I think we **must** have some working defaults for these. @schisamo maybe we can have a read-only account that we can hardcode here so that not every user of mixlib-install is required to set these. If we can't do that I am not sure if requiring every user of mixlib-install to set these is viable.

lib/mixlib/install/backend/artifactory.rb
```diff
@@ -19,72 +19,141 @@
...
+          @client ||= ::Artifactory::Client.new
+        end
 
-          if options.platform
-            params["omnibus.platform"] = options.platform
-            params["omnibus.platform_version"] = options.platform_version
-            params["omnibus.architecture"] = options.architecture
-          end
+        def repository
+          @repository ||= "omnibus-#{options.channel}-local"
```
sersut: I like the fact that you can query other repositories too. :+1:

lib/mixlib/install/options.rb
```diff
@@ -79,6 +78,10 @@ def for_ps1?
         platform == "windows" || shell_type == :ps1
       end
 
+      def latest_version?
+        [:latest, "latest"].include?(product_version)
```
sersut: `product_version.to_sym == :latest`

spec/mixlib/install/backend_spec.rb
```diff
@@ -43,7 +43,11 @@
   shared_examples_for "the right artifact info" do
     it "gives the right url artifact info" do
       if !expected_info.key?(:url)
-        expect(info.url).to match "https://"
+        if channel == :unstable
```
sersut: This is not ideal. Maybe we can set the `expected_info[:url]` in `:unstable` channel tests and not introduce this logic.

lib/mixlib/install/backend/base.rb
```diff
@@ -0,0 +1,20 @@
+module Mixlib
+  class Install
+    class Backend
+      class Base
```
sersut: Why? I would leave this to the future. Introduction of well documented and defined APIs for `mixlib-install` should be done in one go. Let's not do this for now.

lib/mixlib/install/backend/artifactory.rb
schisamo: So the only people using the unstable channel will be Chef Software, Inc. employees and systems (as artifactory.chef.co is internal). Every CI/CD system will be authenticating with a service account  and any employee can use their AD creds.

lib/mixlib/install/generator/powershell.rb
```diff
@@ -65,14 +65,26 @@ def ps1_modularize(module_body, module_name)
...
+                     artifacts: artifactory_backend.info)
+        end
+
+        def artifactory_backend
+          @artifactory_backend ||= Mixlib::Install::Backend::Artifactory.new(options)
+        end
+
+        def product_version
+          if options.for_artifactory?
+            artifactory_backend.product_version
```
sersut: It looks like we are introducing a way to resolve `:latest` to a product version. This should be an API on the `Backend` classes. Is this not breaking `Bourne`?

lib/mixlib/install/backend/artifactory.rb
schisamo: We should be passing these config values to `Artifactory::Client` via an explicit values in the options hash though...see:
https://github.com/chef-cookbooks/artifactory-pro/blob/master/libraries/_helper.rb#L66-L73

patrick-wright: Sure, but where are those values read from?

lib/mixlib/install/generator/powershell.rb
patrick-wright: I didn't like this either.  It does however call the Backend::Artifactory method, I just don't like that it must be called twice.  Does not break for bourne.

patrick-wright:
These changes burned me out today :)  I ran into a few last minute snags.  Is the current channel supposed to store non-integration builds (12.6.0)?  I don't recall seeing those until recently.

I've added `TODOs ` in the specs to show what I mean.

The biggest sore for me is the artifactory versioning handling in `Generator::Powershell`

As of now, all specs and acceptance tests pass.

README.md
```diff
@@ -35,9 +35,18 @@ artifacts.first.url
 # => "http://opscode-omnibus-packages-current.s3.amazonaws.com/mac_os_x/10.9/x86_64/chef-12.5.1%2B20151009083009-1.dmg"
 ```
 
+## Unstable channel
+The `:unstable` channel is currently only available when connected to Chef's internal network.
+Configure Artifactory access by setting the following environment variables:
+```
+# default endpoint
+# export ARTIFACTORY_ENDPOINT='http://artifactory.chef.co'
```
sersut: I would remove this line :smile: to keep things simple. We can help put the one who wants to speak to a different artifactory instance...

lib/mixlib/install/backend/artifactory.rb
```diff
@@ -19,72 +19,114 @@
...
+        #
         def info
-          begin
-            results = client.get("/api/search/prop", params, headers)["results"]
-          rescue Errno::ETIMEDOUT => e
-            raise e, "unstable channel uses endpoint #{ARTIFACTORY_ENDPOINT} \
-which is currently only accessible through Chef's internal network."
-          end
+          artifacts = artifactory_info.collect { |a| create_artifact(a) }
+          artifacts_for_version = artifacts.find_all { |a| a.version == resolve_version(artifacts) }
```
sersut: Neat! :+1:

lib/mixlib/install/backend/artifactory.rb
```diff
@@ -19,72 +19,114 @@
...
-            "omnibus.version" => options.product_version
-          }
+          results = client.post("#{::Artifactory.endpoint}/api/search/aql",
+                                query, "Content-Type" => "text/plain")
 
-          if options.platform
-            params["omnibus.platform"] = options.platform
-            params["omnibus.platform_version"] = options.platform_version
-            params["omnibus.architecture"] = options.architecture
+          results["results"].collect do |result|
```
sersut: Maybe add a comment here about the fact that we construct the `downloadUri` ourselves in order not to make a second call to artifactory.

lib/mixlib/install/generator/powershell.rb
```diff
@@ -65,14 +65,29 @@ def ps1_modularize(module_body, module_name)
...
+          get_script("get_project_metadata_for_artifactory.ps1",
+                     artifacts: artifactory_backend.info)
+        end
+
+        def artifactory_backend
+          @artifactory_backend ||= Mixlib::Install::Backend::Artifactory.new(options)
+        end
+
+        def product_version
+          if options.for_artifactory?
```
sersut: To make this better I would move version resolution logic into options @patrick-wright. i.e. 

```
# options.rb
def resolve_version(artifacts)
  if latest?
    @resolved_version ||= begin
      # ...
      # logic to resolve the version
      #
    else
      @resolved_version ||= version
    end
end

# powershell.rb

def artifactory_urls
  artifacts = artifactory_backend.info
  # Above line resolves the version and resolved version is cached in options
  # ^^ Comment is important because we need to call info before doing anything else :)
  
  get_script("get_project_metadata_for_artifactory.ps1",
                  artifacts: artifacts,
                  resolved_version: options.resolved_version)
  # change the template to use `resolved_version` in it.

# Also use options.resolve_version in backend/artifactory.rb
```

Hopefully makes sense. If not we can chat :smile:

spec/mixlib/install/backend_spec.rb
```diff
@@ -297,21 +301,24 @@
 
         context "with a major.minor product version" do
           let(:product_version) { "12.4" }
-          let(:expected_version) { /^12.4.\d\+[0-9]{14}$/ }
+          # TODO: should current contain non-integration builds? let(:expected_version) { /^12.4.\d\+[0-9]{14}$/ }
```
sersut: Not sure what's up with this. Let's chat with Seth and see what we can do.

lib/mixlib/install/backend/artifactory.rb
patrick-wright: Will do.  I'l going to spend a little time today to see if I can have AQL return the full downloadUri in the results.

PR #37, opened 2015-12-19
remove unnecessary memoization

patrick-wright:
@sersut just wanted to make this change before I forget after the long break


sersut:
:+1:

PR #38, opened 2015-12-29
install on zone host *only* on solaris (sans -G installs in all zones…

curiositycasualty:
… ಠ_ಠ)

/cc @chef/engineering-services @chef/solaris @chef/client-solaris


scotthain:
Yeah that's a good call. :+1:

lamont-granquist:
:+1:

PR #39, opened 2015-12-29
Add retry option to curl

ccope:
wget retries 20 times by default, but curl does not retry at all. This should make the two methods have similar behavior.

### DIFF
--- a/lib/mixlib/install/generator/bourne/scripts/helpers.sh
+++ b/lib/mixlib/install/generator/bourne/scripts/helpers.sh
@@ -136,7 +136,7 @@ do_wget() {
 # do_curl URL FILENAME
 do_curl() {
   echo "trying curl..."
-  curl -sL -D $tmp_dir/stderr "$1" > "$2"
+  curl --retry 5 -sL -D $tmp_dir/stderr "$1" > "$2"
   rc=$?
   # check for 404
   grep "404 Not Found" $tmp_dir/stderr 2>&1 >/dev/null


scotthain:
If I recall correctly, curl does exponential backoff, how many minutes does this end up being? I'm fine with the retry, just not sure about how long it wants to take by default.

ccope:
It does exponentially back off. Each retry doubles the backoff, starting at 1 second and maxing out at 10 minutes (source: https://github.com/bagder/curl/blob/b4a39491cacde68c9ed5499db7fcc873976e1879/docs/curl.1#L1541-L1546 )
20 retries would take excessively long with the default backoff algorithm. What's a reasonable amount of time to spend retrying? If we go with 1 minute, I could cut it down to 5 retries or set a fixed delay of 3 seconds.

scotthain:
I think 1 minute (5 retries) sounds reasonable, anyone else have an opinion?

jtimberman:
My only concern is whether the version of curl on all the platforms we support has this option.

ccope:
curl has had this option since October 2004 https://github.com/bagder/curl/commit/96cf615e9dec951b2c4244780b5f3fe2fb303f5b

scotthain:
@jtimberman we bundle curl 7.36.0 in the build toolchain on our CI builders/testers on some of the more 'interesting' platforms.

ccope:
This gets used in the omnitruck install.sh script, so that would be before anything chef-related was installed.

ccope:
Looking at the Tier 1 supported platforms, the questionable ones appear to be AIX and Solaris 10.
From googling, I don't see any evidence that curl comes pre-installed on AIX, and a couple of people complaining that it lacks curl and wget.
Solaris 10 comes with wget, and not curl.
The next oldest supported OS, RHEL/Centos/Oracle 5, has curl 7.15

jtimberman:
I think it's *probably* OK, as it *should* fall back to perl on aix/solaris. I don't have a good way to verify, though.

patrick-wright:
I am :+1: on this change.  Valid points, but let's see if this fixes the immediate issue and then put our heads together if this proves to be failure point for aix and solaris.  My guess is most likely no since curl won't be the defacto package used in those instances.

schisamo:
:+1: from me. I think we just need @thommay to weigh in.

thommay:
As a point of order, I think this repo should grow a maintainers file and then be handled under RFC 30 - ie, two maintainer +1s is all that's required. 

In this specific instance, :+1:

PR #46, opened 2016-01-26
Use /builds endpoint to determine the latest version when using :unstable

sersut:
When we are running tests in CI, generally there are more than 1 parallel builds happening. When we do not use the build record we pick up the information about the latest build which most of the time is partial. Also when using `items.find()` in AQL artifactory returns a maximum of 1000 items and sometimes it doesn't include the artifacts we are looking for (We have a total of ~ 23k artifacts, way more items)

Still need to: 
- [x] Update specs for `:unstable` channel.

Sharing this for preview and incase there are any initial comments. 

/cc: @patrick-wright, @schisamo

### DIFF
--- a/lib/mixlib/install/backend/artifactory.rb
+++ b/lib/mixlib/install/backend/artifactory.rb
@@ -27,6 +27,7 @@ module Mixlib
       class Artifactory
         class ConnectionError < StandardError; end
         class AuthenticationError < StandardError; end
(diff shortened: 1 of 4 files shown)


schisamo:
:+1:

patrick-wright:
per our convo:
- raise an exception when the build api doesn't return any artifacts for the product
- add comments around artifactory 3.0 limitations to joining the /build get and the aql

PR #52, opened 2016-02-04
update acceptance suites

patrick-wright:
@sersut 

1) fixes `kitchen` exec => `bundle exec kitchen`
2) allows `KITCHEN_INSTANCES=ubuntu bundle exec chef-acceptance test` to just run ubuntu suites

I think we should  consider removing Gemfile.lock
Note Gemfile is referencing a local chef-acceptance path


acceptance/Gemfile
```diff
@@ -2,7 +2,8 @@ source "https://rubygems.org"
 
 gem "mixlib-install", path: "../"
 gem "test-kitchen", github: "sersut/test-kitchen", branch: "sersut/mixlib-install-update"
-gem "chef-acceptance", github: "chef/chef-acceptance"
+# gem "chef-acceptance", github: "chef/chef-acceptance"
+gem "chef-acceptance", path: "../../chef-acceptance"
```
sersut: We should revert this.

acceptance/current/.kitchen.yml
```diff
@@ -1,9 +1,9 @@
 driver:
   name: vagrant
-  forward_agent: yes
```
sersut: Any reason we're removing `forward_agent` ?

sersut:
Yeah let's remove Gemfile.lock :smile:

acceptance/current/.kitchen.yml
patrick-wright: I thought that was just needed for vbox issues for some reason.  We can put it back.

PR #53, opened 2016-02-04
Give users a way to get 32 / 64 bit packages on Windows

jaym:
I've added a switch that lets people force the architecture. If the switch is not specified, it will default to `auto`, which will give 64 bit packages on the current channel if the machine supports it.

### DIFF
--- a/lib/mixlib/install/generator/powershell/scripts/get_project_metadata.ps1.erb
+++ b/lib/mixlib/install/generator/powershell/scripts/get_project_metadata.ps1.erb
@@ -66,7 +66,7 @@ function Get-ProjectMetadata {
   Write-Verbose "Platform Version: $platform_version"
 
   if ($architecture -eq 'auto') {
-    if (!($env:ProgramW6432 -eq $null) -And ($channel -eq 'current')) {
+    if (((get-wmiobject win32_operatingsystem).osarchitecture -like '64-bit') -and ($channel -like 'current')) {
       $architecture = 'x86_64'
     } else {
       $architecture = 'i386'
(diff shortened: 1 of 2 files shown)


patrick-wright:
:+1:

jaym:
Going to test on 2008 before i merge

schisamo:
:+1:

jaym:
I think the accidental purge broke the tests

jaym:
@adamedx @sersut what do we want to do when people ask for stable + 64 bit

Perhaps a warning saying it could change to an actual 64 bit package?

adamedx:
I don't think it needs to do anything at this point -- there will only be a short window of time where it gives you a 32-bit package when you ask for 64, and today's omnitruck api always gives you 32-bit if you ask for 64, so it's no worse than that experience.

lib/mixlib/install/generator/powershell/scripts/get_project_metadata.ps1.erb
```diff
@@ -62,16 +65,22 @@ function Get-ProjectMetadata {
   }
   Write-Verbose "Platform Version: $platform_version"
 
-  # TODO: When we ship a 64 bit ruby for Windows.
-  $machine = 'x86_64'
-  Write-Verbose "Machine: $machine"
+  if ($architecture -eq 'auto') {
+    if (!($env:ProgramW6432 -eq $null) -And ($channel -eq 'current')) {
```
smurawski: Rather than relying on the environmental variable, I'd use something like 

```powershell
if ((get-wmiobject win32_operatingsystem).osarchitecture -like '64-bit') -and ($channel -like 'current')){
```

smurawski:
Other than the method of OS architecture detection :+1:

PR #55, opened 2016-02-09
Remove tests that rely on us not purging the current channel

jaym:
I've deleted the tests that rely on certain packages
being in the current channel. This was broken by
the great purge.


jaym:
Fixed in #56

PR #56, opened 2016-02-09
Fix current specs by picking a version dynamically from omnitruck.

sersut:
I think this is slightly better @jaym. Still not perfect but at least we are exercising the code path to call / parse `/versions` endpoint.

Replaces https://github.com/chef/mixlib-install/pull/55

### DIFF
--- a/spec/mixlib/install/backend_spec.rb
+++ b/spec/mixlib/install/backend_spec.rb
@@ -42,6 +42,31 @@ context "Mixlib::Install::Backend" do
 
   let(:expected_protocol) { "https://" }
 
+  # We need full version number for some of our tests and
+  # for :current and :stable channels we auto-purge artifacts.
+  # So we query omnitruck here to find a version so that we can
+  # run our tests
+  def pick_version_for(channel)
+    Mixlib::Install.new(
+      channel: channel,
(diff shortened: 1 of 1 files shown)


jaym:
Yea, I went down the path of implementing this, however stopped when I realized the `/versions` only returns the latest. The issue with using the latest is that you won't be able to distinguish between a bug that accidentally always fetches the latest and the correct behavior of fetching the thing you asked for.

That being said, I'm :+1: for this change understanding why it's not perfect. Some suggested modifications if you want to do them:

- The `artifact_info` method on `Mixlib::Install` returns a list of `ArtifactInfo`, which have a `to_hash` you could use for expected_info (perhaps merging your expectation for version). This would require modifying the platform, platform version, and architecture to use that of the  `ArtifactInfo` you have.
- You might be able to get away with calling `artifact_info` once (may speed up the tests)

If you don't have time to do those / don't think they're valuable, :shipit: as this is better than what we have now.

sersut:
@jaym you're totally right @jaym. Modifying the p, pv and m will be valuable but it will also complicate the tests a lot and cost / benefit is against our favor IMO. 

I'll say let's :shipit: and unblock ourselves for now.

PR #57, opened 2016-02-10
add arch support for unstable channel

patrick-wright:
Fixes windows architecture selection when using unstable channel (artifactory)

tested locally for chef-client


patrick-wright:
@sersut @chefsalim

sersut:
@patrick-wright nice catch! :+1: 

Can we run `rake unstable`? Some of the tests should be broken there since we expect specific versions and they are not available due to auto-cleanup in artifactory. We should fix them with a similar change like https://github.com/chef/mixlib-install/pull/56.

tyler-ball:
:+1:

PR #58, opened 2016-02-10
fix unstable specs, update rake tasks, update travis

patrick-wright:
@sersut


sersut:
:+1: All good. Tried to fail the spec we were skipping but was not able to.

I vote to not skip it or if you have a repro I'm curious :smile: 

Also pushed a change to remove the skipping.

PR #62, opened 2016-03-08
Add chef-server-ha-provisioning package to the list

irvingpop:
I noticed that `chef-server-ha-provisioning` was missing from the product matrix, and therefore not consumable by chef-ingredient.   This PR should fix that.

Note to reviewers:  There's two additional commits I added to this PR.  One to fix failing style checks and the second to fix a failing spec test.


rhass:
I am :+1: as long as we don't cause any weirdness as the default for mixlib-install is to check the stable channel for available packages.

yzl:
:+1:

sersut:
:+1:

PR #63, opened 2016-03-08
detect platform info using api

patrick-wright:
Adds `self.detect_platform` method to `Mixlib::Install`.

Calling the method will run the platform_detection script and parse the output in order to return a Hash of platform info.

This allows chef-ingredient to merge platform properties to be passed to the Mixlib::Install initialization.

### DIFF
--- a/lib/mixlib/install.rb
+++ b/lib/mixlib/install.rb
@@ -100,6 +100,44 @@ module Mixlib
     end
 
     #
+    # Automatically set the platform options
+    #
+    def detect_platform
+      options.set_platform_info(self.class.detect_platform)
+    end
+
+    #
+    # Returns a Hash containing the platform info options
+    #
+    def self.detect_platform
+      platform_info = if RbConfig::CONFIG["host_os"] =~ /mswin|mingw/
+                        `#{self.detect_platform_ps1}`
+                      else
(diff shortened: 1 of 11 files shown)


patrick-wright:
@sersut This is ready for initial review

lib/mixlib/install.rb
```diff
@@ -100,6 +100,23 @@ def upgrade_available?
     end
 
     #
+    # Returns a Hash containing the platform info options
+    #
+    def self.detect_platform(shell_type = :sh)
```
sersut: I think we need to expose these to the outside world in two ways:

1. We need to make `detect_platform()` an instance method that updates platform, platform_version and architecture for the `options` of the instance. The usage of it from chef-ingredient (or any other tool) will be like:
```ruby
installer = Mixlib::Install.new({
  product_name: new_resource.product_name,
  channel: new_resource.channel,
  product_version: new_resource.version
}).detect_platform
```
2. We need to have `self.detect_platform_sh` and `self.detect_platform_ps1` which returns the script to run incase a tool wants to do this remotely on a different node.

sersut:
One comment about the apis we are introducing. Rest is :trophy:.

sersut:
:+1:

PR #67, opened 2016-03-14
Add chef-compliance product support

alexpop:
Added `chef-compliance` to SUPPORTED_PRODUCT_NAMES.
Had to change the name used by the `product` resource as Artifactory was 404ing with just `compliance`.

All these examples worked with the `pw/mixlib-install-source` branch of `chef-ingredient`:

```
chef_ingredient 'chef-compliance' do
  channel :current
  artifactory_username node['artifactory']['username']
  artifactory_password node['artifactory']['password']
  version :latest
  action [:install, :reconfigure]
end
```

```
chef_ingredient 'chef-compliance' do
  channel :unstable
  artifactory_username node['artifactory']['username']
  artifactory_password node['artifactory']['password']
  version :latest
  action [:install, :reconfigure]
end
```

```
chef_ingredient 'chef-compliance' do
  artifactory_username node['artifactory']['username']
  artifactory_password node['artifactory']['password']
  version :latest
  action [:install, :reconfigure]
end
```

### DIFF
--- a/lib/mixlib/install/options.rb
+++ b/lib/mixlib/install/options.rb
@@ -36,6 +36,7 @@ module Mixlib
         chef
         chefdk
         chef-server
+        chef-compliance
         delivery-cli
         omnibus-toolchain
         push-jobs-client
--- a/lib/mixlib/install/product.rb
+++ b/lib/mixlib/install/product.rb
@@ -215,7 +215,7 @@ PRODUCT_MATRIX = Mixlib::Install::ProductMatrix.new do
     package_name "chefdk"
   end
 
-  product "compliance" do
+  product "chef-compliance" do
     product_name "Chef Compliance"
     package_name "chef-compliance"
     ctl_command "chef-compliance-ctl"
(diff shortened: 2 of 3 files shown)


patrick-wright:
:+1: cc/ @chef/engineering-services

schisamo:
@patrick-wright So I think the fact that `compliance` was used to query Artifactory is just a bug on how we wired things into the Artifactory backend. The package_name is set to `chef-compliance` and this should be used when querying Artifactory. I believe we still want people using a `compliance` as the product key just like we use `analytics`, and `manage` for other products. Package names change over time. The only products that should be prefixed with `chef-*` are `chef-server` and potentially `chef-backend` if/when that ships.

patrick-wright:
@schisamo Thanks for the info.
@alexpop This PR will not meet your requirements.  Update the product to look like this:
```ruby
  product "compliance" do
    product_name "Chef Compliance"
    package_name "chef-compliance"
    ctl_command "chef-compliance-ctl"
    config_file "/etc/chef-compliance/chef-compliance.rb"
  end
```
This will allow `chef-ingredient "compliance"` and will still query the correct property in artifactory.

sersut:
@alexpop We have added support for compliance in a different PR. We should not be needing this anymore.

PR #69, opened 2016-03-18
Bintray backend for mixlib-install

sersut:
Add Bintray backend.  Can be implemented directly, not yet wired into `Mixlib::Install` api.
* Init sinatra test server (will continue other PRs for artifactory, and possibly omnitruck)


lib/mixlib/install/backend/bintray.rb
```diff
@@ -0,0 +1,151 @@
...
+  class Install
+    class Backend
+      class Bintray < Base
+        ENDPOINT = "https://bintray.com/api/v1".freeze
+
+        API_VERSION = "v1".freeze
+
+        API_SUBJECT = "chef".freeze
+
+        DEFAULT_USERNAME = "mixlib-install@chef".freeze
```
sersut: I think it would be good to rename these to `BINTRAY_USERNAME` & `BINTRAY_PASSWORD`.

lib/mixlib/install/backend/bintray.rb
```diff
@@ -0,0 +1,151 @@
...
+
+module Mixlib
+  class Install
+    class Backend
+      class Bintray < Base
+        ENDPOINT = "https://bintray.com/api/v1".freeze
+
+        API_VERSION = "v1".freeze
+
+        API_SUBJECT = "chef".freeze
```
sersut: `API_SUBJECT` ? Is that a bintray term?

lib/mixlib/install/backend/bintray.rb
```diff
@@ -0,0 +1,151 @@
...
+          @client ||= Mixlib::Install::Bintray::Client.new(
+            endpoint: endpoint,
+            username: ENV.fetch("BINTRAY_USERNAME", DEFAULT_USERNAME),
+            password: ENV.fetch("BINTRAY_PASSWORD", DEFAULT_PASSWORD)
+          )
+        end
+
+        #
+        # Get latest version for product/channel
+        #
```
sersut: Would be good to capture what is returned here with `@return`.

lib/mixlib/install/backend/bintray.rb
```diff
@@ -0,0 +1,151 @@
...
+          # Delete .asc files
+          results.each { |r| results.delete(r) if r["name"].end_with?(".asc") }
+
+          # Convert results to build records
+          results.map { |a| create_artifact(a) }
+        end
+
+        #
+        # Creates an instance of ArtifactInfo
+        #
```
sersut: It would be good to have an example of how `artifact_map` looks like.

lib/mixlib/install/backend/bintray.rb
```diff
@@ -0,0 +1,151 @@
...
+          # platform/platform_version/filename
+          path = artifact_map["path"].split("/")
+
+          platform = path[0]
+
+          platform_version = path[1]
+
+          filename = artifact_map["name"]
+
+          architecture = if %w{ x86_64 amd64 x64 }.fuzzy_include?(filename)
```
sersut: Let's move this block into its own function.

lib/mixlib/install/backend/bintray.rb
```diff
@@ -0,0 +1,210 @@
...
+  end
+end
+
+module Mixlib
+  class Install
+    class Backend
+      class Bintray < Base
+        ENDPOINT = "https://bintray.com".freeze
+
+        API_VERSION = "v1".freeze
```
schisamo: I think we should simplify things and include the version as part of the endpoint so:

```ruby
ENDPOINT = "https://bintray.com/api/v1".freeze
```

In REST API terminology the "endpoint" can include a path (`/api/v1`) along with the scheme (`https`) and host (`bintray.com`).

lib/mixlib/install/backend/bintray.rb
```diff
@@ -0,0 +1,210 @@
...
+
+        BINTRAY_ORGANIZATION = "chef".freeze
+
+        BINTRAY_USERNAME = "mixlib-install@chef".freeze
```
schisamo: It might be worthwhile to throw a comment above this username/password explaining that this is intentional and not an accidental leaking of credentials.

lib/mixlib/install/backend/bintray.rb
```diff
@@ -0,0 +1,210 @@
...
+        BINTRAY_PASSWORD = "a83d3a2ffad50eb9a2230f281a2e19b70fe0db2d".freeze
+
+        DOWNLOAD_URL_ENDPOINT = "https://packages.chef.io".freeze
+
+        def endpoint
+          @endpoint ||= ENV.fetch("BINTRAY_ENDPOINT", ENDPOINT)
+        end
+
+        def org
+          @org ||= ENV.fetch("BINTRAY_ORGANIZATION", BINTRAY_ORGANIZATION)
```
schisamo: I don't see a reason for this to be configurable. We will only ever support the `chef` org, no need to give people the allusion we support anything else.

lib/mixlib/install/backend/bintray.rb
```diff
@@ -0,0 +1,210 @@
...
+                         elsif %w{ i386 x86 }.fuzzy_include?(filename)
+                           "i386"
+                         elsif %w{ powerpc }.fuzzy_include?(filename)
+                           "powerpc"
+                         elsif %w{ sparc }.fuzzy_include?(filename)
+                           "sparc"
+                         elsif platform == "mac_os_x"
+                           "x86_64"
+                         else
+                           "unknown"
```
schisamo: Would it be better to raise here with an `UnknownArchitecture` exception or similar?

lib/mixlib/install/backend/bintray.rb
```diff
@@ -0,0 +1,210 @@
...
+
+        private
+
+        #
+        # Create Bintray client
+        #
+        # @return [Mixlib::Install::Bintray::Client] Mixlib::Install::Bintray::Client instance
+        #
+        def client
+          @client ||= Mixlib::Install::Bintray::Client.new(
```
schisamo: Where is `Mixlib::Install::Bintray::Client` defined? I don't see it in this PR and I also don't see a require for it at the top of this file.

spec/mixlib/install/backend/bintray_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "spec_helper"
+require "mixlib/install/backend/bintray"
+
+context "Mixlib::Install::Backend::Bintray" do
+  use_bintray_test_server = true
```
sersut: I'm not a fan of this style. I think using `before(:all)` here will make things more readable.

sersut:
A minor comment but otherwise :+1: 

Let's make sure the commits are squashed with a informing commit message before merge.

PR #71, opened 2016-03-21
allow architecture option to be passed to powershell installer

mwrock:
currently the powershell install command will, by default, set the architecture of the product to install based on the architecture of the installing node. Thats reasonable but the `render_command` method has no way of passing an architecture in to `Install-Product`. This adds that.

Note that it makes no sense to do this to the bourne shell script because it does not use a dynamic `architecture` variable.

One side effect of this change and where I may be making incorrect assumptions is that it removes the validation ensuring that `platform`, `platform_version`, and `architecture` are all passed together. This needs to be able to pass architecture without passing in the others and I did not see the value for changing it to remove architecture and thus removed it all together. So feel free to comment on why that should at least partially be put back.

### DIFF
--- a/lib/mixlib/install/generator/powershell.rb
+++ b/lib/mixlib/install/generator/powershell.rb
@@ -89,11 +89,11 @@ module Mixlib
         end
 
         def render_command
-          <<EOS
-install -project #{options.product_name} \
--version #{product_version} \
--channel #{options.channel}
-EOS
+          cmd = "install -project #{options.product_name}"
+          cmd << " -version #{product_version}"
+          cmd << " -channel #{options.channel}"
+          cmd << " -architecture #{options.architecture}" if options.architecture
+          cmd << "\n"
         end
       end
     end
--- a/lib/mixlib/install/options.rb
+++ b/lib/mixlib/install/options.rb
@@ -70,7 +70,6 @@ module Mixlib
 
         errors << validate_product_names
         errors << validate_channels
-        errors << validate_platform_info
         errors << validate_shell_type
 
         unless errors.compact.empty?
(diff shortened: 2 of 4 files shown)


mwrock:
cc @sersut @patrick-wright

adamedx:
Adding @tyler-ball.

lib/mixlib/install/generator/powershell.rb
```diff
@@ -92,7 +92,8 @@ def render_command
           <<EOS
 install -project #{options.product_name} \
 -version #{product_version} \
--channel #{options.channel}
+-channel #{options.channel} \
+-architecture #{options.architecture || 'auto'}
```
sersut: Can we only add `-architecture #{options.architecture}` to the command when architecture exists rather than adding `"auto"` all the times? Here is what I'm thinking of:
```
install_command += "-architecture #{options.architecture}" if options.architecture

sersut:
@patrick-wright and I thought about what's the best way to handle this and this sounds looks like the best approach @mwrock. But we do not want to add `"auto"` as a possible value of the architecture. Since powershell scripts are using `"auto"` when architecture does not exist I think things are going to work fine. 

One of the things we need to do is to add a `full_platform_info_present?` method which will check for everything being present when we need it since we are removing a validation but that's OK not to have it in this PR. We will add it in the future when we're touching mixlib-install.

patrick-wright:
:+1: after architecture changes

mwrock:
thanks for the comments all. @sersut I only add architecture now if the option is given. I'll merge after your +1

sersut:
:+1:

PR #73, opened 2016-03-23
Expose private bintray_get method to the world.

sersut:

This PR goes with https://github.com/chef/omnitruck/pull/168 and exposes the `bintray_get` method to the outside world. We need this in omnitruck while polling bintray.

### DIFF
--- a/lib/mixlib/install/backend/bintray.rb
+++ b/lib/mixlib/install/backend/bintray.rb
@@ -16,7 +16,9 @@
 # limitations under the License.
 #
 
+require "json"
 require "mixlib/install/backend/base"
+require "mixlib/install/artifact_info"
 
 #
 # Add method to Array class to support
@@ -71,7 +73,29 @@ module Mixlib
           artifacts.length == 1 ? artifacts.first : artifacts
         end
 
-        private
+        #
+        # Makes a GET request to bintray for the given path.
+        #
+        # @param [String] path
+        #   "/api/v1/packages/chef" is prepended to the given path.
+        #
+        # @return [String] JSON parsed string of the bintray response
+        #
+        def bintray_get(path)
+          uri = URI.parse(endpoint)
+          http = Net::HTTP.new(uri.host, uri.port)
+          http.use_ssl = (uri.scheme == "https")
+
+          full_path = File.join(uri.path, "packages/chef", path)
+          request = Net::HTTP::Get.new(full_path)
+          request.basic_auth(BINTRAY_USERNAME, BINTRAY_PASSWORD)
+
+          res = http.request(request)
+
+          # Raise if response is not 2XX
+          res.value
+          JSON.parse(res.body)
+        end
 
         #
         # Get latest version for product/channel
@@ -79,7 +103,7 @@ module Mixlib
         # @return [String] latest version value
         #
         def latest_version
-          result = bintray_get("versions/_latest")
+          result = bintray_get("#{options.channel}/#{options.product_name}/versions/_latest")
           result["name"]
         end
 
@@ -90,10 +114,16 @@ module Mixlib
         #
         def bintray_artifacts
           version = options.product_version == :latest ? latest_version : options.product_version
-          results = bintray_get("versions/#{version}/files")
+          results = bintray_get("#{options.channel}/#{options.product_name}/versions/#{version}/files")
 
+          #
           # Delete .asc files
-          results.each { |r| results.delete(r) if r["name"].end_with?(".asc") }
+          # Also delete .pkg files which are uploaded along with dmg files for
+          # some chef versions.
+          #
+          results.reject! do |r|
+            r["name"].end_with?(".asc") || r["name"].end_with?(".pkg")
+          end
 
           # Convert results to build records
           results.map { |a| create_artifact(a) }
@@ -150,20 +180,7 @@ module Mixlib
           platform_version = path[1]
 
           filename = artifact_map["name"]
-          architecture = if %w{ x86_64 amd64 x64 }.fuzzy_include?(filename)
-                           "x86_64"
-                         elsif %w{ i386 x86 }.fuzzy_include?(filename)
-                           "i386"
-                         elsif %w{ powerpc }.fuzzy_include?(filename)
-                           "powerpc"
-                         elsif %w{ sparc }.fuzzy_include?(filename)
-                           "sparc"
-                         elsif platform == "mac_os_x"
-                           "x86_64"
-                         else
-                           raise UnknownArchitecture,
-                                 "architecture can not be determined for '#{filename}'"
-                         end
+          architecture = parse_architecture_from_file_name(filename)
 
           {
             platform: platform,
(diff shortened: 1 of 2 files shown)


lib/mixlib/install/backend/bintray.rb
```diff
@@ -152,14 +178,16 @@ def parse_platform_info(artifact_map)
...
+                         elsif %w{ powerpc sun4u solaris2.5.10 solaris2.5.9 }.fuzzy_include?(filename)
                            "powerpc"
-                         elsif %w{ sparc }.fuzzy_include?(filename)
+                         elsif %w{ sparc sun4v }.fuzzy_include?(filename)
                            "sparc"
                          elsif platform == "mac_os_x"
                            "x86_64"
+                         elsif %w{ msi }.fuzzy_include?(filename)
+                           "i386"
                          else
```
yzl: This is totally my ignorance, but do the `ppc64` and `ppc64le` architectures need to be represented in this list?

lib/mixlib/install/backend/bintray.rb
```diff
@@ -152,14 +178,16 @@ def parse_platform_info(artifact_map)
           filename = artifact_map["name"]
           architecture = if %w{ x86_64 amd64 x64 }.fuzzy_include?(filename)
                            "x86_64"
-                         elsif %w{ i386 x86 }.fuzzy_include?(filename)
+                         elsif %w{ i386 x86 i86pc i686 }.fuzzy_include?(filename)
                            "i386"
-                         elsif %w{ powerpc }.fuzzy_include?(filename)
+                         elsif %w{ powerpc sun4u solaris2.5.10 solaris2.5.9 }.fuzzy_include?(filename)
```
schisamo: There is no such thing as Solaris with `powerpc` arch so this should remain:

```ruby
elsif %w{ powerpc }.fuzzy_include?(filename)
```
further down you need to make the sparc section:

```ruby
elsif %w{ sparc sun4u sun4v solaris2.5.9 }.fuzzy_include?(filename)
```

`solaris2.5.10` is actually `i386` so it should be in that final section:

```ruby
elsif %w{ msi solaris2.5.10 }.fuzzy_include?(filename)
```

In the future you can use https://github.com/sethvargo/chef-sugar/blob/master/lib/chef/sugar/architecture.rb as a guide along with the `omnibus.*` properties attached to published artifacts in artifactory.chef.co.

scotthain:
@schisamo is correct re: SPARC/sun4u/sun4v, and I believe @yzl is also correct about the ppc64le/ppc64. I'm a little confused about the `solaris2.5.X` - can we chat about this change?

schisamo:
@scotthain The `solaris2.5.X` bit is because in the past the `*.solaris` packages did not include the architecture in the name. Since Bintray does not have freeform metadata we can attach the actual artifacts we have to infer the architecture for the package names.

lib/mixlib/install/backend/bintray.rb
sersut: @yzl yeah. Adding them right now.

scotthain:
:+1:

lib/mixlib/install/backend/bintray.rb
```diff
@@ -172,20 +191,71 @@ def parse_platform_info(artifact_map)
...
+        # We determine the architecture  based on the filename of the artifact
+        # since architecture the artifact is published for is not available
+        # in bintray.
+        #
+        # @param [String] filename
+        #
+        # @return [String]
+        #   one of the standardized architectures for Chef packages:
+        #   x86_64, i386, powerpc, sparc, ppc64, ppc64le
+        def parse_architecture_from_file_name(filename)
```
schisamo: Do we want to add some spec coverage for this method....seems like something we could break easily.

lib/mixlib/install/backend/bintray.rb
```diff
@@ -71,6 +73,30 @@ def info
...
+        #   "/api/v1/packages/chef" is prepended to the given path.
+        #
+        # @return [String] JSON parsed string of the bintray response
+        #
+        def bintray_get(path)
+          uri = URI.parse(endpoint)
+          http = Net::HTTP.new(uri.host, uri.port)
+          http.use_ssl = (uri.scheme == "https")
+
+          full_path = File.join(uri.path, "packages/chef", path)
```
schisamo: IMO we should not inject `packages/chef` into the provided path. It should just be passed in with the path the caller of this method provides.

sersut: Reason I moved that here is because otherwise we will need to include it pretty much all other places. I think we can do this refactor when we are going towards an `bintray-client` gem.

lib/mixlib/install/backend/bintray.rb
sersut: The best test here is to run omnitruck's poller and make sure there are no errors. I can add some basic things here but they will be most copy paste the test cases here to the spec. 

How about adding some notes saying make sure that omnitruck poller is happy if you are changing this method?

lib/mixlib/install/backend/bintray.rb
yzl: Would there be any gain to doing `def bintray_get(path, subject = 'chef')` now?  Or is it cleaner to let this be until the `bintray-client` refactor?

yzl:
:+1:

lib/mixlib/install/backend/bintray.rb
schisamo: Let's definitely add the comment but I think some spec coverage would be prudent here. These are the types of methods where unit tests make tons of sense. This reminds me of the `Omnibus::Metadata` class:
https://github.com/chef/omnibus/blob/master/lib/omnibus/metadata.rb
https://github.com/chef/omnibus/blob/master/spec/unit/metadata_spec.rb

sersut: Cool :+1:

lib/mixlib/install/backend/bintray.rb
sersut: I think we can leave out for now. I don't think that will help with anything right now.

spec/mixlib/install/backend/bintray_spec.rb
```diff
@@ -105,4 +106,47 @@
...
+          "chef_12.8.1-1_i86pc.deb",
+          "chef_12.8.1-1.msi",
+        ],
+        "powerpc" => [
+          "chef-12.8.1-1.powerpc.bff",
+        ],
+        "sparc" => [
+          "chef-12.8.1-1.sparc.solaris",
+          "chef-12.8.1-1.sun4u.solaris",
+          "chef-12.8.1-1.sun4v.solaris",
```
schisamo: Let's add a few of the legacy `solaris2.5.9` artifacts and a few `solaris2.5.10` in the `i386` section.

schisamo:
@sersut One more small comment, otherwise :+1: :shipit:

spec/mixlib/install/backend/bintray_spec.rb
sersut: Ah nice catch! Totally missed that :smile:

PR #74, opened 2016-03-25
Load the Bintray backend with `Mixlib::Install::Backend`

schisamo:
This allows us to instantiate an instance of the Bintray backend after a simple `require 'mixlib/install'`.

/cc @sersut @patrick-wright


sersut:
:+1:

PR #75, opened 2016-03-26
add a blank version value for non specific versions on windows just like linux

mwrock:
this uses the same version url value as we use for linux [here](https://github.com/chef/mixlib-install/blob/master/lib/mixlib/install/script_generator.rb#L110).

### DIFF
--- a/lib/mixlib/install/script_generator.rb
+++ b/lib/mixlib/install/script_generator.rb
@@ -69,7 +69,7 @@ module Mixlib
                               sudo_command}
 
       def initialize(version, powershell = false, opts = {})
-        @version = version || "latest"
+        @version = (version || "latest").to_s.downcase
         @powershell = powershell
         @http_proxy = nil
         @https_proxy = nil
@@ -198,7 +198,7 @@ module Mixlib
 
         url = "#{base}#{endpoint}"
         url << "?p=windows&m=x86_64&pv=2008r2" # same package for all versions
-        url << "&v=#{CGI.escape(version.to_s.downcase)}"
+        url << "&v=#{CGI.escape(version)}" unless %w{latest true nightlies}.include?(version)
         url << "&prerelease=true" if prerelease
         url << "&nightlies=true" if nightlies
         url


lib/mixlib/install/script_generator.rb
```diff
@@ -195,10 +195,11 @@ def windows_metadata_url
         base = if omnibus_url =~ %r{/install.sh$}
                  "#{File.dirname(omnibus_url)}/"
                end
+        msi_version = %w{latest true nightlies}.include?(version) ? "" : version
```
jkeiser: Did you intend for the boolean `true` to work here, or just the string?

lib/mixlib/install/script_generator.rb
```diff
@@ -195,10 +195,11 @@ def windows_metadata_url
...
 
         url = "#{base}#{endpoint}"
         url << "?p=windows&m=x86_64&pv=2008r2" # same package for all versions
-        url << "&v=#{CGI.escape(version.to_s.downcase)}"
+        url << "&v=#{CGI.escape(msi_version.to_s.downcase)}"
```
jkeiser: Would this go cleaner if you just append `unless %w{latest true nightlies}.include?(version)` to this line? Then no need for new variable or indirection, and you don't pass a parameter you don't mean to pass.

mwrock: yeah. that would be. I'll make that change.

lib/mixlib/install/script_generator.rb
mwrock: expecting a string like https://github.com/chef/mixlib-install/blob/windows_fix/lib/mixlib/install/script_generator.rb#L110

patrick-wright:
@thommay Since you have the most context with this code could review please?

schisamo:
We should perform a `to_s` on any version passed in via the initializer here:
https://github.com/chef/mixlib-install/blob/2ec4d1b398bf3234c5eae2ee18891b07ec3d75d7/lib/mixlib/install/script_generator.rb#L72

so something like this:

```ruby
@version = version.to_s || "latest"
```

This will ensure symbols are properly compared as strings. It also saves having to `to_s` the version everywhere else.

mwrock:
made the change to the initializer

PR #76, opened 2016-03-28
Remove md5 and sha1 usage from install.sh.

sersut:
This PR removes the usage of md5 and sha1 in `install.sh`. With the recent migration to bintray, we do not keep track of the md5 for our packages anymore. We had a conversation with @schisamo and @lamont-granquist in Slack and decided that removing sha1 and md5 is not going to lead to common breakages. 

We also wanted to add a new check using `openssl` to cover some of the historical platforms that does not have sha256 support. 

### DIFF
--- a/lib/mixlib/install/generator/bourne/scripts/artifactory_urls.sh.erb
+++ b/lib/mixlib/install/generator/bourne/scripts/artifactory_urls.sh.erb
@@ -12,7 +12,6 @@
 # Outputs:
 # $download_url:
 # $sha256:
-# $md5:
 ############
 
 <% artifacts.each do |artifact| %>
@@ -20,7 +19,6 @@
   mkdir -p $artifact_info_dir
   touch "$artifact_info_dir/artifact_info"
   echo "url <%= artifact.url%>" >> "$artifact_info_dir/artifact_info"
-  echo "md5 <%= artifact.md5%>" >> "$artifact_info_dir/artifact_info"
   echo "sha256 <%= artifact.sha256%>" >> "$artifact_info_dir/artifact_info"
 <% end %>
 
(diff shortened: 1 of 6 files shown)


sersut:
@scotthain @lamont-granquist @schisamo So far I have tested this on:

* mac
* solaris-10-sun4v
* aix-6-powerpc

I am having this issue with openssl on Solaris 10:
```
downloaded metadata file looks valid...
downloading https://packages.chef.io/stable/solaris2/5.10/chef-12.8.1-1.sparc.solaris
  to file /tmp/install.sh.1821/chef-12.8.1-1.sparc.solaris
trying wget...
trying perl...
trying python...
Comparing checksum with openssl...
unknown option '-sha256'
options are
-c              to output the digest with separating colons
-d              to output debug info
-hex            output as hex dump
-binary         output in binary form
-sign   file    sign digest using private key in file
-verify file    verify a signature using public key in file
-prverify file  verify a signature using private key in file
-keyform arg    key file format (PEM or ENGINE)
-signature file signature to verify
-binary         output in binary form
-engine e       use engine e, possibly a hardware device.
-md5 to use the md5 message digest algorithm (default)
-md4 to use the md4 message digest algorithm
-sha1 to use the sha1 message digest algorithm
-sha to use the sha message digest algorithm
-mdc2 to use the mdc2 message digest algorithm
-ripemd160 to use the ripemd160 message digest algorithm
Package checksum mismatch!
```

I think the older versions of openssl does not have this flag. Shall we:

1. Remove `openssl` check all together?
2. Improve it and make it more resilient?

lamont-granquist:
yeah for historical reference centos5, solaris 11, aix 7 have both md5 and shasum that supports sha256.

sol10 and aix 6 both lack all of the above and have neither md5 nor sha256, so are :sadpanda: no matter what.

i believe i dropped md5 in there for old pre-10.9 MacOSX which had md5 but lacked any shasum--which we no longer support.

lamont-granquist:
i don't think the openssl check actually gets us any more O/S distros.  i'd be fine with #1.  you could wrap the check (probably similarly to what we do with error checking around the various wget/curl/perl/python/fetch/etc methods of downloading) so that if it throws a non-zero exit code it doesn't consider that failed.

lamont-granquist:
and md5 and sha1 are also broken now for this purpose.

sersut:
@lamont-granquist @schisamo 

This is now good to go. I have removed `openssl` since it doesn't make us gain more coverage.

lamont-granquist:
:+1:

schisamo:
:shipit:

PR #77, opened 2016-03-28
Fix Product Keys

schisamo:
See individual commit messages for full details.

/cc @sersut @patrick-wright

### DIFF
--- a/lib/mixlib/install/product.rb
+++ b/lib/mixlib/install/product.rb
@@ -172,19 +172,6 @@ PRODUCT_MATRIX = Mixlib::Install::ProductMatrix.new do
     config_file "/etc/chef-backend/chef-backend.rb"
   end
 
-  product "chef-ha" do
-    product_name "Chef Server High Availability addon"
-    package_name "chef-ha"
-    config_file "/etc/opscode/chef-server.rb"
-  end
-
-  product "chef-marketplace" do
-    product_name "Chef Cloud Marketplace addon"
-    package_name "chef-marketplace"
-    ctl_command "chef-marketplace-ctl"
-    config_file "/etc/chef-marketplace/marketplace.rb"
-  end
-
   product "chef-server" do
     product_name "Chef Server"
     package_name do |v|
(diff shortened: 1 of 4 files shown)


sersut:
@schisamo the code is good. But I think this will be a compat-breaking change. I am especially concerned for the users of `chef-ingredient`.

IMO, we need to either add some code (and deprecation messages) in chef-ingredient or here.

lib/mixlib/install/product.rb
```diff
@@ -198,18 +185,6 @@ def lookup(key, version = :latest)
     config_file "/etc/opscode/chef-server.rb"
   end
 
-  product "chef-server-ha-provisioning" do
```
patrick-wright: Has this been deprecated?  I don't know anything about it, but I know @irvingpop added it recently.

patrick-wright:
:+1:

lib/mixlib/install/product.rb
schisamo: I didn't see any builds of this product in either current or stable:
https://bintray.com/chef/current
https://bintray.com/chef/stable

It could be it hasn't been built in a while and has been reaped from current. Can you provide some context @irvingpop?

schisamo:
@sersut Re: compat breaking I would bump chef-ingredients major version and get people using the correct product keys. Longer term that will be more humane for end users than keeping a bunch of weird compat code around. Thoughts?

sersut:
@schisamo I like this change but I don't think is a super big change to introduce a major version bump to chef-ingredient. I think we should add some deprecation messages and ad the old names into chef-ingredient. 

Major version bumps should be cheap but I think it will be more delightful to give users some time with deprecation messages.

lib/mixlib/install/product.rb
irvingpop: @patrick-wright @schisamo quite the opposite, this is an EA product.  I don't know if a build has gone through since the cutover from packagecloud.   @charlesjohnson is the best person to talk to.

schisamo:
@sersut Then I would add the deprecation messages and compat you are seeking in chef-ingredient not here in mixlib-install.

sersut:
@schisamo sounds good. We can merge this once we resolve `chef-server-ha-provisioning`. @patrick-wright and I can add the deprecation messages in chef-ingredient. We will be looking into that next.

lib/mixlib/install/product.rb
charlesjohnson: It's not dead. It just hasn't needed a build in a while. We should get it back into current if it's been reaped out.

sersut:
:+1:

PR #79, opened 2016-03-29
release 0.8.0

mwrock:


jkeiser:
:+1:

sersut:
IMO we should be bumping the version to `1.0.0` since there are major API differences between `0.7.X` and `master`. @thommay @schisamo thoughts?

Also we would like to merge https://github.com/chef/mixlib-install/pull/82 in before making the release @mwrock. Can we wait until that is merged? /cc: @patrick-wright

patrick-wright:
@mwrock If the features you need are in the latest 0.8.0 alpha release then I would say let's release that as 0.8.0.  If what you need is still floating in master I suppose we could still release as 0.8.0, but I would like to take a look first if that's the case. 

Could you help us understand why you need a release right now? (perhaps the test-kitchen brach merge?)

PR #82 would certainly make sense to release as 1.0.0.

mwrock:
Thanks for looking @patrick-wright. Currently acceptance tests in chef and chef-dk leverage master of this gem and @sersut 's branch but that is now two and soon to be 3 releases behind the released test-kitchen. So we miss the intent of chef-dk acceptance tests testing what it actually ships which is tk 1.6 (probably 1.7 next dk). I really just want to be able to run the chef-acceptance tests with the released tk.

mwrock:
Since the last alpha is just 2 days old, I'd say thats fine to release as 0.8.0. I sould not have my fix merged yesterday but we'd just keep the "hack" in tk to work around it.

patrick-wright:
Ok.  I'm not opposed to releasing 0.8.0 today.  Unless anyone tells me not to -- it shall be done.

sersut:
I definitely think that we should get https://github.com/chef/mixlib-install/pull/82 in before we do a release. I do not think we should have two rels of mixlib-install majorly different than each other. 

Let's wait for that PR to get merged. It will not be a big delay anyways.

patrick-wright:
The test-kitchen branch they are using is 3 releases behind.  Also, I am not sure exactly how long it's going to take to finish #82.  This is a 0. versioned product, so I don't feel we are breaking an semver contracts. Anyone who needs a 0.7 release can lock it down.  I don't want to impede releasing code because we are making deep cuts.  IMHO this should have been released much sooner.

mwrock:
#82 looks good but if its gonna take long to finish off, I agree with @patrick-wright that an 0. release is perfectly legit. There may be a tk 1.7 this week and I don't want to miss that train. Nothing in #82 that tk currently depends on.

mwrock:
closing this since it sounds like a 1.0 is imminent.

PR #85, opened 2016-03-31
adding stopaction to kick in the catch statement if manifest is missing

vinyar:
@mwrock


mwrock:
:+1: 
@vinyar and I looked at this tonight and found that earlier versions of chef (11.14.2 in our case) that lack a `version-manifest.txt`, do not invoke the fallback chef-solo -v to obtain the version. This is because the error output by the `Get-Content` cmdlet on a missing file is a "non terminanting" error. In otherwords it will not terminate the pipeline and trigger a `catch`. Instead it simply outputs an error record t othe pipeline and continues to process it.

Adding `-ErrorAction stop` forces a terminating error and will cause the `catch` to execute.

PR #86, opened 2016-03-31
wait for msiexec to exit

mwrock:
So you'd think that calling `start-process` with `-wait` would prevent the cmdlet returning until `msiexec` completes. In most cases it does. I have seen cases in the past where this is not the case. Oddly this works as expected on my vagrant boxes but I was investigating https://github.com/test-kitchen/kitchen-ec2/issues/241 and was able to consistently reproduce on an EC2 instance calling `start-process -wait` returns immediately and the returned `Process` instance has a blank `ExitCode`.

This forces it to block and fixed EC2. I don't know why EC2 behaves differently here but I do know I have seen this before and have had to use a similar method to work around the problem.


zl4bv:
Seen something like this before (in EC2), so fascinated to consider that it might be an EC2-specific behaviour.

patrick-wright:
:+1:

PR #87, opened 2016-03-31
replace install_command.ps1 md5 with sha256

patrick-wright:
install_command.ps1 still referenced md5 checks.  This caused issues when using the ScriptGenerator.

There was a reference in helpers.ps1 in the generator scripts what was being bypassed, but I just removed it altogether.

@sersut @schisamo @mwrock

### DIFF
--- a/lib/mixlib/install/generator/bourne/scripts/helpers.sh
+++ b/lib/mixlib/install/generator/bourne/scripts/helpers.sh
@@ -210,10 +210,6 @@ do_checksum() {
     echo "Comparing checksum with sha256sum..."
     checksum=`sha256sum $1 | awk '{ print $1 }'`
     return `test "x$checksum" = "x$2"`
-  elif exists shasum; then
-    echo "Comparing checksum with shasum..."
-    checksum=`shasum -a 256 $1 | awk '{ print $1 }'`
-    return `test "x$checksum" = "x$2"`
   else
     echo "WARNING: could not find a valid checksum program, pre-install shasum or sha256sum in your O/S image to get valdation..."
     return 0
--- a/lib/mixlib/install/generator/bourne/scripts/helpers.sh
+++ b/lib/mixlib/install/generator/bourne/scripts/helpers.sh
@@ -210,6 +210,10 @@ do_checksum() {
     echo "Comparing checksum with sha256sum..."
     checksum=`sha256sum $1 | awk '{ print $1 }'`
     return `test "x$checksum" = "x$2"`
+  elif exists shasum; then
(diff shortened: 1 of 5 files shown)


schisamo:
Overall this looks good but don't we want the Windows logic to operate like the Unix logic in that it tries `SHA256` and falls back to `SHA1`:
https://github.com/chef/mixlib-install/blob/b2495ce9db896ce8c9c0444282e67da9d5a62a7b/lib/mixlib/install/generator/bourne/scripts/helpers.sh#L208-L221

We should be consistent across the board.

lib/mixlib/install/generator/bourne/scripts/helpers.sh
```diff
@@ -210,10 +210,6 @@ do_checksum() {
     echo "Comparing checksum with sha256sum..."
     checksum=`sha256sum $1 | awk '{ print $1 }'`
     return `test "x$checksum" = "x$2"`
-  elif exists shasum; then
-    echo "Comparing checksum with shasum..."
-    checksum=`shasum -a 256 $1 | awk '{ print $1 }'`
-    return `test "x$checksum" = "x$2"`
```
schisamo: Actually I misread this code @patrick-wright. It wasn't doing `SHA1` checks but using `shasum` to validate the `SHA256`. We def want to leave this clause in there for compatibility on Unixes that might now have the `sha256sum` binary installed.

mwrock:
The changes look ok. I'm gonna test on a win test run.

lib/mixlib/install/generator/powershell/scripts/helpers.ps1
```diff
@@ -60,18 +60,9 @@ function Test-ProjectPackage {
...
-            $hash.Hash = ([BitConverter]::ToString($c.ComputeHash($in))).Replace("-", "").ToUpper()
-          }
-        }
-      }
-      elseif ($Algorithm -like 'SHA256') {
-        use ($c = New-Object -TypeName Security.Cryptography.SHA256CryptoServiceProvider) {
-          use ($in = (gi $path).OpenRead()) {
-            $hash.Hash = ([BitConverter]::ToString($c.ComputeHash($in))).Replace("-", "").ToUpper()
-          }
+      use ($c = New-Object -TypeName Security.Cryptography.SHA256CryptoServiceProvider) {
```
mwrock: unrelated, but I really like this `use` pattern to replicate a c# like `using`

mwrock:
:+1: pending @schisamo 's comment

this checks out on windows kitchen runs both using `product_name` and `require_chef_omnibus` paths.

PR #89, opened 2016-03-31
Use 32-bit artifacts for 64-bit on windows when we do not have native 64-bit packages available.

sersut:
/cc: @patrick-wright 

Without this change when architecture is `x86_64` mixlib-install can not find any packages for `chefdk` since all `chefdk` builds are published for 32 bit only.


patrick-wright:
:+1:

PR #95, opened 2016-04-06
use .net 2 compliant sha 256 class so we can install on windows 2k8r2

mwrock:
Currently installing on Windows 2008 R2 because `SHA256CryptoServiceProvider` is only available in .net 3.5 and higher which is not the default runtime on Windows 2008 R2. We can use `SHA256Managed` which has a equivilent API but it is not FIPS 140-2 validated and will break if the node is running in fips mode. This should be fine since only chef-client needs FIPS compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## [1.0.4]
+### Fixed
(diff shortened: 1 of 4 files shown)


sersut:
Don't know about the contents of the change but that's all the places we need to change in mixlib-install. Also we need to `bundle update mixlib-install` in omnitruck if we want to update `install.ps1` that gets served over there. 

:+1:

scotthain:
👍 Thanks!

PR #98, opened 2016-04-12
filter metadata.json files from artifactory queries

patrick-wright:
Ensure `metadata.json` files are excluded from artifactory queries

@chef/engineering-services @sersut @tyler-ball @chefsalim


spec/mixlib/install/artifactory_spec.rb
```diff
@@ -56,4 +59,10 @@
     end
   end
 
+  context "when an querying artifact", focus: true do
```
sersut: Forgot `focus`.

schisamo: Need to remove that `focus: true`.

schisamo:
:+1: once specs are green

PR #107, opened 2016-04-25
Backwards compatibility option for platform_version.

sersut:
As a user, I would like to be able to consume Chef packages on new platforms that Chef Software Inc. did not have a chance to add to its official support matrix.

As distro release new versions, it takes us some time to add them into our official support matrix which requires adding new testers and builders to our CI infrastructure. Sometimes we do this in a timely manner but even these cases we have some adventurous users who would like to try Chef on these new versions before they become publicly available.

To satisfy these scenarios this PR adds in an opt-in functionality to mixlib-install and chef-ingredient which lets users get the packages built and tested for the latest version if we do not have their specified version in our official support matrix.

As a note, we have always had this functionality previously under the name "yolo".

- [x] Chat with @schisamo and test this in the currently broken delivery-cluster scenario.
- [x] Investigate if we need to change the name of this new option. If you have any new suggestions please drop it in as a comment. A few guidelines:
  - we do not want to use abbreviations that are hard to understand like `yolo` or `pv`.
  - we do not want to make the name scary like `allow_unstable` or `allow_unsafe`.
  - we do not want to make the name generic since we might introduce more options for backwards compatibility in the future outside platform_version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,24 @@ artifact.platform # => "mac_os_x"
 artifact.platform_version # => "10.10"
 ```
 
+### Use an artifact released for an earlier version of the platform
+```ruby
+options = {
+  channel: :current,
+  product_name: 'chef',
+  product_version: :latest,
+  platform: 'ubuntu',
+  platform_version: '15.04',
+  architecture: 'x86_64',
+  platform_version_compatibility_mode: true
+}
+
+artifact = Mixlib::Install.new(options).artifact_info
+
+artifact.platform # => "ubuntu"
+artifact.platform_version # => "14.04"
+```
+
 ## Unstable channel
 The `:unstable` channel is currently only available when connected to Chef's internal network.
 
--- a/lib/mixlib/install/backend/artifactory.rb
+++ b/lib/mixlib/install/backend/artifactory.rb
@@ -42,25 +42,12 @@ module Mixlib
         #
         # @return [Array<ArtifactInfo>] list of artifacts for the configured
         # channel, product name, and product version.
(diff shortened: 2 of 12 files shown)


scotthain:
👍 code looks good. I like the proposed name.

yzl:
👍

mmzyk:
Am I reading this correctly that the proposed name is platform_version_backward_compat_mode? It's not a bad name, but it confuses me slightly, because I think about backward compatibility as being for old versions, yet part of this functionality is being able to install a package on a potentially newer platform. I see how it makes sense if you consider it from the package and not the OS. 

Ideas that I've typed out to see how they worked:
platform_version_fall_back_mode
platform_version_mistmatch_mode
platform_version_compatability_mode

I actually like platform_version_compatability_mode. Is this too generic to use (it might be now that I've read the requirements again)? If so, the original proposition is fine, as I have nothing better to offer. 

Based on the constraints that y'all understand better than me, pick the option that makes the most sense and let's roll forward. I don't want to be a bottle neck in this (and probably already have been).

scotthain:
+1 on `platform_version_compatibility_mode`

schisamo:
I'm also down with `platform_version_compatability_mode`. It's a really long option name but with the constraints we've set for ourselves it is what it is.

scotthain:
<pedant> let's make sure we spell it correctly </pedant>

mmzyk:
@scotthain Yeah, don't go by my spelling. I'm a terrible speller. :)

I'm 👍 to all this.

PR #110, opened 2016-05-03
Added support for retrying msi installation for exit code 1618

gshively11:
Chef provisioning would constantly fail to install the chef client msi on our openstack windows nodes due to another msi installation already being in progress.  Added some simple retry code to detect exit code 1618 (another install in progress) and retry the installation.  Retries 10 times, although it usually works after 1 or 2.  Verified this works using latest chefdk (13.21).

### DIFF
--- a/support/install_command.ps1
+++ b/support/install_command.ps1
@@ -43,11 +43,25 @@ Function Download-Chef($url, $sha256, $dst) {
 
 Function Install-Chef($msi) {
   Log "Installing Chef Omnibus package $msi"
-  $p = Start-Process -FilePath "msiexec.exe" -ArgumentList "/qn /i $msi" -Passthru -Wait
-  $p.WaitForExit()
-
-  if ($p.ExitCode -ne 0) { throw "msiexec was not successful. Received exit code $($p.ExitCode)" }
-
+  $installingChef = $True
+  $installAttempts = 0
+  $maxAttempts = 10
+  while ($installingChef) {
+    $installAttempts++;
+    $p = Start-Process -FilePath "msiexec.exe" -ArgumentList "/qn /i $msi" -Passthru -Wait
+    $p.WaitForExit()
+    if ($p.ExitCode -ne 0) {
+      if ($installAttempts -gt $maxAttempts) {
+        throw "msiexec was not successful, exceeded retry limit (10).  Received exit code $($p.ExitCode)"
+      } elseif ($p.ExitCode -ne 1618) {
+        throw "msiexec was not successful. Received exit code $($p.ExitCode)"
+      } else {
+        Log "Another msi install is in progress (exit code 1618), retrying..."
+        continue
+      }
+    }
+    $installingChef = $False
+  }
   Remove-Item $msi -Force
   Log "Installation complete"
 }
--- a/support/install_command.ps1
+++ b/support/install_command.ps1
@@ -44,7 +44,6 @@ Function Download-Chef($url, $sha256, $dst) {
 Function Install-Chef($msi) {
   Log "Installing Chef Omnibus package $msi"
   $installingChef = $True
-  $installAttempts = 0
   while ($installingChef) {
     $installAttempts++
     $p = Start-Process -FilePath "msiexec.exe" -ArgumentList "/qn /i $msi" -Passthru -Wait
--- a/support/install_command.ps1
+++ b/support/install_command.ps1
@@ -44,6 +44,7 @@ Function Download-Chef($url, $sha256, $dst) {
 Function Install-Chef($msi) {
   Log "Installing Chef Omnibus package $msi"
   $installingChef = $True
+  $installAttempts = 0
   while ($installingChef) {
     $installAttempts++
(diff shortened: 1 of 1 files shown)


chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.getchef.com/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

## GitHub Users Who Are Not Authorized To Contribute

The following GitHub users do not appear to have signed a CLA:

* @gshively11

[Please sign the CLA here.](https://supermarket.chef.io/become-a-contributor)

chef-supermarket:
Hi. Your friendly Curry bot here. Just letting you know that all commit authors have become authorized to contribute. I have added the "Signed CLA" label to this issue so it can easily be found in the future.

sersut:
Thanks for sending this in @gshively11. Also let's ping @ksubrama @btm. Once we have green light from our windows experts we can ship this.

mwrock:
I've definitely run into provisioning scenarios where MSI's fight each other. Taking the retry course as you have, I would use fewer retries and inject a `start-sleep 10` for each retry. Another option to consider is use a function like this (I stole this from code I have used for years):
```
function Wait-ForMSIEXEC{
    Write-Host "Checking for other running MSIEXEC installers..." -Verbose
    Do{
        Get-Process | ? {$_.Name -eq "MSIEXEC"} | % {
            if(!($_.HasExited)){
                $proc=Get-WmiObject -Class Win32_Process -Filter "ProcessID=$($_.Id)"
                if($proc.CommandLine -ne $null -and $proc.CommandLine.EndsWith(" /V")){ break }
                Write-Host "Another installer is running: $($proc.CommandLine). Waiting for it to complete..."
                $_.WaitForExit()
            }
        }
    } Until ((Get-Process | ? {$_.Name -eq "MSIEXEC"} ) -eq $null)
}
```
This just waits for the running MSI to finish and it reports the current MSI being installed and that would help to troubleshoot long running MSIs.

I'm good with either implementation, the above or adding a delay to your retries.

gshively11:
@mwrock I like the latter, will test that out here in a few minutes

gshively11:
@mwrock, copied that snippet and called the func right before the install, still got the 1618.  likely a race condition.  I'll opt for fewer retries with a delay, commit here in a few

gshively11:
@mwrock I tried doing a 10 second sleep with 3 retries and it failed each time with 1618.  I have a feeling there are a bunch of installs queued up, so the 10-retries-with-no-delay install is able to sneak its way in, but the sleep introduces too many chances for another msi to take over.  Any other recommendations?

gshively11:
Haha, I just had it fail after reverting back to the 10 attempts with no delay.  Guess I need to take another crack at this.

mwrock:
computers can be so cruel @gshively11

gshively11:
Is there ever a case where we wouldn't want to keep retrying the chef-client install in the event of a 1618?  If the alternative is failing the provisioning run, it seems better to just keep waiting.  The only downside I can see is if some MSI process zombies and provisioning just waits forever.  Maybe have it wait forever, but have it start logging warning messages to the console about things taking longer than expected?

mwrock:
I agree. Just keep trying. That removes complexity around retry counting. There is already a timeout wrapping the entire chef-client run so I'm not worried about an indefinite hang.

gshively11:
@mwrock how's that look?

mwrock:
looks good. Might get rid of the `$installAttempts` variable since that really isn't doing much for ya but I wouldn't hold that against this. :+1:

gshively11:
I removed it, but then added it back in because the next provision run was scrolling those wait messages off screen and I was curious how many attempts it actually took :)

support/install_command.ps1
```diff
@@ -43,11 +43,20 @@ Function Download-Chef($url, $sha256, $dst) {
...
-
+  $installingChef = $True
+  $installAttempts = 0
+  while ($installingChef) {
+    $installAttempts++
+    $p = Start-Process -FilePath "msiexec.exe" -ArgumentList "/qn /i $msi" -Passthru -Wait
+    $p.WaitForExit()
+    if ($p.ExitCode -eq 1618) {
+      Log "Another msi install is in progress (exit code 1618), retrying ($($installAttempts))..."
+      continue
```
sersut: Do we want to put in some `sleep` in here in order not to poll constantly?

gshively11: I found it to be more effective without the sleep.  My assumption is that there are multiple MSI installs queued up somewhere else.  Constantly attempting to install gives you a better chance of slipping between installs.  When I added the sleep it took longer for it to finally trigger the install.

gshively11: As it stands, it already takes a few seconds between installs, since it takes the MSI a second to register that another install is in progress.

sersut: 👍

PR #111, opened 2016-05-12
Remove project validation in `install_project.ps1`

schisamo:
This brings the generated PowerShell script into parity with the Bourne version which does not do any project-level validation. It also allows some additional projects to be properly installed on Windows:

* delivery-cli
* push-jobs-client

### DIFF
--- a/lib/mixlib/install/generator/powershell/scripts/install_project.ps1
+++ b/lib/mixlib/install/generator/powershell/scripts/install_project.ps1
@@ -16,11 +16,6 @@ function Install-Project {
   [cmdletbinding(SupportsShouldProcess=$true)]
   param (
     # Project to install
-    # chef - Chef Client
-    # chefdk - Chef Development Kit
-    # angrychef - AngryChef
-    # server and container are not valid windows targets
-    [validateset('chef', 'chefdk', 'angrychef')]
     [string]
     $project = 'chef',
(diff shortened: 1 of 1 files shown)


lib/mixlib/install/generator/powershell/scripts/install_project.ps1
```diff
@@ -16,11 +16,6 @@ function Install-Project {
   [cmdletbinding(SupportsShouldProcess=$true)]
   param (
     # Project to install
-    # chef - Chef Client
-    # chefdk - Chef Development Kit
-    # angrychef - AngryChef
-    # server and container are not valid windows targets
-    [validateset('chef', 'chefdk', 'angrychef')]
     [string]
     $project = 'chef',
```
rhass: With the other projects in mind, shouldn't the channels apply to them as well and not just `'chef'`?

schisamo: Not sure I follow @rhass. All line 20 does is set the default value of project to `chef`...like we do in the Bourne version here:
https://github.com/chef/mixlib-install/blob/v1.0.11/lib/mixlib/install/generator/bourne/scripts/script_cli_parameters.sh#L16

scotthain:
👍

PR #115, opened 2016-06-03
Platform Architecture Detection on stable

smurawski:
Previously, since we only published i386 builds on omnitruck stable, there was logic to prevent autodetection of the platform architecture.

For the most part omnitruck metadata does the right thing if the real architecture is passed to stable.  Also, we have x86_64 builds of chef-client and deliver-cli in stable which could now be picked up by default (and i386 systems will just not see them).

Also added a bit more error handling around `new-uri` to stop the commands if there is an invalid uri, rather than showing an error and continuing to try to download.

### DIFF
--- a/lib/mixlib/install/generator/powershell/scripts/get_project_metadata.ps1.erb
+++ b/lib/mixlib/install/generator/powershell/scripts/get_project_metadata.ps1.erb
@@ -53,13 +53,8 @@ function Get-ProjectMetadata {
   $platform_version = Get-PlatformVersion
   Write-Verbose "Platform Version: $platform_version"
 
-  # Custom architecture detection based on channel
   if ($architecture -eq 'auto') {
-    if (((get-wmiobject win32_operatingsystem).osarchitecture -like '64-bit') -and ($channel -like 'current')) {
-      $architecture = 'x86_64'
-    } else {
-      $architecture = 'i386'
-    }
+    $architecture = Get-PlatformArchitecture
   }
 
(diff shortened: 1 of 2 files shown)


schisamo:
This is definitely the correct fix but I'm worried that after we update Omnitruck to use a version of mixlib-install that contains this change ChefDK won't get properly installed on 64-bit Windows instances that use Omnitruck's `install.ps1` (we only build/publish 32-bit ChefDK packages). We'll probably need to tweak Omnitruck with a change similar to:
https://github.com/chef/omnitruck/commit/888a9bc956e4323aa587cc05e4f25069d4704c55

I believe @sersut is in the process of pulling some of this platform/version/arch logic up into mixlib-install so I'll want him to weigh in on things also.

smurawski:
@schisamo The omnitruck metadata endpoint returns the right thing for chefdk when you have 'x86_64' as the architecture

```
> Get-ProjectMetadata -project chefdk -architecture x86_64 | fl *

Name  : url
Key   : url
Value : https://packages.chef.io/stable/windows/2012r2/chefdk-0.14.25-1-x86.msi

Name  : sha256
Key   : sha256
Value : b2ce916f6c30909e16d7df444d6812bd84fc44cf8868d4f2b9e1c5fed5878f50

Name  : version
Key   : version
Value : 0.14.25

Name  : sha1
Key   : sha1
Value : 4fd902d826a4073556afe5e9c1fab7015f035507
```

PR #118, opened 2016-06-27
[JEX-178] add env var that forces artifactory as the backend for all channels

patrick-wright:
This will be the first PR of two for the upcoming Artifactory backend change.  This PR can be merged now. Setting `MIXLIB_INSTALL_BACKEND=artifactory` will enable the new behavior. Without setting the env var mixlib-install will behave as before.

Added `spec/mixlib/install/backend/artifactory_all_channels_spec.rb` which is basically a copy of the Bintray backend spec. It was updated to maintain parity when using Artifactory for stable and current channels.

@sersut @schisamo @rhass @yzl


schisamo:
Should we add a quick unit test?

sersut:
Code-wise this is OK. I think we need to spend some time to think about a test plan for this change and also to form a foundation for our overall test plan. Let's chat more about it in standup..

patrick-wright:
I can add a test.  Also keep in mind this is an interim change. The permanent changes are in this branch: https://github.com/chef/mixlib-install/compare/pw/jex-178-2

mixlib-install.gemspec
```diff
@@ -25,6 +25,8 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "rspec", "~> 3.3"
   spec.add_development_dependency "pry"
+  spec.add_development_dependency "rb-readline"
   spec.add_development_dependency "vcr"
   spec.add_development_dependency "webmock", "~> 1.0"
+  spec.add_development_dependency "chefstyle", "~> 0.3"
```
sersut: Out of curiosity, was there a specific reason for why we pinned `chefstyle` instead of using master?

sersut:
Overall 👍 Can we identify a high level test plan for this change and cover some of it before hitting `merge`?

sersut:
👍 Let's squash and :shipit: this.

PR #119, opened 2016-06-27
Add omnibus project property to Product mapping

jkeiser:
Artifactory backend now uses omnibus_project name when querying artifactory repo.


schisamo:
This looks like the correct change. In most all cases the Omnibus project name and the package name are the same....except for Chef Server!
https://github.com/chef/chef-server/blob/master/omnibus/config/projects/chef-server.rb#L17-L23

lib/mixlib/install/product.rb
```diff
@@ -181,45 +187,52 @@ def lookup(key, version = :latest)
...
+    omnibus_project "chef-ha"
+    config_file     "/etc/opscode/chef-server.rb"
   end
 
   product "manage" do
     product_name "Management Console"
     package_name do |v|
       v < version_for("2.0.0") ? "opscode-manage" : "chef-manage"
     end
+    omnibus_project "manage"
```
schisamo: This should actually follow the same version logic and naming as the package name:
https://github.com/chef/chef-manage/blob/master/omnibus/config/projects/chef-manage.rb#L17
https://github.com/chef/omnibus-manage/blob/master/config/projects/opscode-manage.rb#L17

lib/mixlib/install/product.rb
```diff
@@ -233,56 +246,64 @@ def lookup(key, version = :latest)
...
+    ctl_command     "private-chef-ctl"
+    config_file     "/etc/opscode/private-chef.rb"
   end
 
   product "push-jobs-client" do
     product_name "Chef Push Server"
+    omnibus_project "push-jobs-client"
     package_name do |v|
       v < version_for("1.3.0") ? "opscode-push-jobs-client" : "push-jobs-client"
     end
```
schisamo: Ditto on here. The omnibus project name changed at the same time as the package name.

lib/mixlib/install/product.rb
```diff
@@ -144,30 +145,35 @@ def lookup(key, version = :latest)
   # Products in alphabetical order
 
   product "analytics" do
-    product_name "Analytics Platform"
-    package_name "opscode-analytics"
-    ctl_command "opscode-analytics-ctl"
-    config_file "/etc/opscode-analytics/opscode-analytics.rb"
+    product_name    "Analytics Platform"
+    package_name    "opscode-analytics"
+    omnibus_project "opscode-analytics"
```
sersut: I think instead of adding `omnibus_project` as a standard property we can do something like below:

```
def omnibus_project(value = nil)
  if value.nil?
    @omnibus_project || @package_name
  else
    @omnibus_project = value
  end
```

lib/mixlib/install/product.rb
```diff
@@ -29,6 +29,7 @@ def initialize(&block)
         :ctl_command,
         :package_name,
         :product_name,
+        :omnibus_project,
```
sersut: Do we still need this here?

sersut:
👍

schisamo:
👍

PR #120, opened 2016-07-07
fix omnibus_project method

patrick-wright:
When `omnibus_project` was set for a product where `package_name` was being passed a block to determine the name based on version, `omnibus_project` was being set to the Proc instance because it was defaulting to the `@package_name` instance variable. This should have been calling the `package_name` method.

The resulting error was:
```
 had an error: URI::InvalidURIError: bad URI(is not URI?): /api/build/#<Proc:0x00000008bff948@/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/mixlib-install-1.1.0/lib/mixlib/install/product.rb:233>
```

@rhass @sersut

### DIFF
--- a/lib/mixlib/install/product.rb
+++ b/lib/mixlib/install/product.rb
@@ -72,7 +72,7 @@ module Mixlib
       #
       def omnibus_project(value = nil)
         if value.nil?
-          @omnibus_project || @package_name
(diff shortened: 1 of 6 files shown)


schisamo:
:+1:

PR #127, opened 2016-08-12
Recognize Arista EOS

jerearista:
Enable detection of Arista EOS platforms using the same string as ohai as discussed with @jjasghar


wrightp:
👍

scotthain:
👍

PR #131, opened 2016-08-19
Non-root provision without sudo privelege

baurzhan-konurbayev:
In some environments chef-client can be executed from non-root user, who doesn't have sudo privilege as well. We have root and non-root provision for application user. So when root downloaded chef-client into /tmp, non-root fails to download it to the same location due to permission issue.

### DIFF
--- a/support/install_command.sh
+++ b/support/install_command.sh
@@ -1,4 +1,10 @@
-tmp_stderr="/tmp/stderr";
+basedir="$HOME/.chef"
+
+if [ ! -d $basedir ]; then
+  mkdir -p $basedir
+fi
+
+tmp_stderr="$basedir/stderr";
 
 # capture_tmp_stderr SOURCE
 capture_tmp_stderr() {
@@ -215,8 +221,8 @@ main() {
       chef_omnibus_url=`echo "$chef_omnibus_url" | sed -e "s/https/http/"`;
     fi
 
-    do_download "$chef_omnibus_url" /tmp/install.sh;
-    $sudo_sh /tmp/install.sh $install_flags;
+    do_download "$chef_omnibus_url" $basedir/install.sh;
+    $sudo_sh $basedir/install.sh $install_flags;
   else
     echo "-----> Chef Omnibus installation detected (${pretty_version})";
   fi


chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.getchef.com/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

## Non-GitHub Verified Committers

There are 1 commit author(s) whose commits are authored by a non-GitHub verified email address. Chef will have to manually verify that they are authorized to contribute.

[Please sign the CLA here.](https://supermarket.chef.io/become-a-contributor)

wrightp:
Can you explain in more detail what sort of priv issue you were having when working in the `tmp` dir?

baurzhan-konurbayev:
Hi!

This may be not broad case, but we have 2 level provisions.
The first one is root provision, which take cares of base packages and
configuration.
The second is application user provision.
We install chef-client for both user, but application user runs is without
sudo priviledge.
So, when root creates and downloads chef-install and other files to /tmp,
our application user is not able to download to overwrite files owned by
root.

Thanks



-- 
С уважением, Бауржан Конурбаев.

wrightp:
OK. I understand your use case. I'm going to ask for a few opinions on this. We may be able to achieve the same result without having to download the same (potential) file in different locations.

wrightp:
@thommay Would you mind putting in your 2 pence on this PR?

thommay:
I don't understand why you need to install chef twice, rather than just having two config files?

baurzhan-konurbayev:
In our production system location of chef-client owned by root is not
accessible by application user at all.
This is historical thing. We are going to change our things work to root
provision only, but it will happen later, because there are a lot of things
(depricated features) to refactor due to moving to chef server 12.8

Thanks.

-- 
С уважением, Бауржан Конурбаев.

PR #132, opened 2016-09-01
package router backend

wrightp:
- PackageRouter is now _the_ backend
- consolidated specs
- generated new fixture data (and now there is just one fixture dir)

- [x] Validate omnitruck
- [x] Validate bourne and ps1 scripts

We could really benefit from some integration tests to validate script generators. Not much code coverage otherwise.


schisamo:
@wrightp We have delivered the versioned `/api` endpoint to packages.chef.io so you should be able to remove the `/etc/hosts` workaround.

lib/mixlib/install/backend/package_router.rb
```diff
@@ -219,56 +193,22 @@ def map_properties(properties)
...
-          rescue ::Artifactory::Error::HTTPError => e
-            if e.code == 401 && e.message =~ /Bad credentials/
-              raise AuthenticationError, "Artifactory server denied credentials."
-            else
-              raise e
-            end
-          end
-
-          results
+          @endpoint ||= ENV.fetch("PACKAGE_ROUTER_ENDPOINT", ENDPOINT)
```
schisamo: We'll need to port some of the old `Mixlib::Install::Backend::Bintray#url` [method](https://github.com/chef/mixlib-install/blob/d04dc39b68847f654d51b7ff1f97fa5806b9b477/lib/mixlib/install/backend/bintray.rb#L166-L171) into this class somewhere, namely the bit that used a non-SNI, non-SSL version for legacy platforms:

```ruby
base_url = case "#{platform_info[:platform]}-#{platform_info[:platform_version]}"
           when "freebsd-9", "el-5", "solaris2-5.9", "solaris2-5.10"
             COMPAT_DOWNLOAD_URL_ENDPOINT
           else
             DOWNLOAD_URL_ENDPOINT
           end
```

The new `COMPAT_DOWNLOAD_URL_ENDPOINT` should be `http://packages.chef.io`.

wrightp: I guess I don't understand this. Won't these platforms automatically redirect to the default endpoint?

schisamo: @wrightp They will not be able to download over the default `https://` endpoint because of OLD SSL. We want to fall back to the `http://` version of the endpoint.

sersut:
Other than @schisamo's point I can not think of anything else. This looks 🚀 🍰 

👍

lib/mixlib/install/backend/package_router.rb
wrightp: :doh: OK that's what I didn't see

lib/mixlib/install/backend/package_router.rb
```diff
@@ -219,56 +195,32 @@ def map_properties(properties)
...
 
         # Generates a chef standard download uri in the form of
-        # http://endpoint/channel/platform/platform_version/filename
-        def generate_chef_standard_path(channel, platform, platform_version, filename)
+        # http://endpoint/files/:channel/:project/:version/:platform/:platform_version/:file
+        def generate_chef_standard_path(channel, project, version, platform, platform_version, filename)
+          # For some older platform & platform_version combinations we need to
+          # use COMPAT_DOWNLOAD_URL_ENDPOINT since these versions have an
+          # OpenSSL version that can not verify the ENDPOINT based urls
+          base_url = case "#{platform}-#{platform_version}"
```
wrightp: @schisamo here's where I'm implemented the COMPAT_DOWNLOAD_URL_ENDPOINT logic.

wrightp: los specs

https://github.com/chef/mixlib-install/blob/fac27e2bcb1fbb28d5f7fa945d9ad2a64bc5cb70/spec/mixlib/install/backend_spec.rb#L51-L62

lib/mixlib/install/backend/package_router.rb
```diff
@@ -195,8 +197,18 @@ def map_properties(properties)
...
         def generate_chef_standard_path(channel, project, version, platform, platform_version, filename)
+          # For some older platform & platform_version combinations we need to
+          # use COMPAT_DOWNLOAD_URL_ENDPOINT since these versions have an
+          # OpenSSL version that can not verify the ENDPOINT based urls
+          base_url = case "#{platform}-#{platform_version}"
+                     when "freebsd-9", "el-5", "solaris2-5.9", "solaris2-5.10"
+                       COMPAT_DOWNLOAD_URL_ENDPOINT
+                     else
+                       endpoint
+                     end
```
schisamo: This logic has to go in `#endpoint` like it did previously in the `Mixlib::Install::Backend::Bintray` or API calls on these legacy platforms will also fail (since `#get` uses `#endpoint` when making the actual API requests).

wrightp: I think the way I have it is correct. We generate the full url path when creating each ArtifactInfo instance. If we match on those select platforms we will use the `compat` endpoint, otherwise we will use the default or overridden endpoint. 

Maybe what we want here is to control the protocal. hostname to `packages.chef.io` and toggle whether to use ssl or not.

wrightp: This endpoint is just for the download URLs as we don't care about direct mixlib-install API calls.

schisamo: But any API requests made from the legacy instance will fail right? For these legacy instances we want both the generated file download URLs and API requests to use `http://packages.chef.io` (vs `https://packages.chef.io`).

wrightp: Perhaps, but this would be a new feature. We only supported ssl and non-sll download url paths. We never handled API requests differently. Do we feel we need to add this now?

schisamo: No but what I am suggesting would exactly match what is being done in the existing `Mixlib::Install::Backend::Bintray` backend so if we want to preserve behavior that seems like the correct path.

PR #133, opened 2016-09-02
WIP: New acceptance updates

wrightp:
New acceptance test suites for validating install.sh and install.ps1 scripts

PR #134, opened 2016-09-06
Add support for Nano

mwrock:
This PR adds support for installing chef on Windows Nano. It two key code adjustments:

1. Detect if running on the `.net core` runtime and use the appropriate APIs. Most of this is just namespace changes but downloading internet content is a pretty big change.
2. If installing a file with a `.appx` extension, use the appx packaging mechanisms for detection, install and upgrade of the client.

A few things to note:
* Non-nano versions of windows should behave without any change to current logic
* There is no proxy support for nano yet. That should be addressed but I had issues getting `IWebProxy` properly implemented and am putting this to the side for now.
* The appx packager installs all packages to a common "application store" on disk. This is problematic because the path contains spaces and is also non deterministic. Therefore we create a symbolic link pointing the `omnibus_install_root` path to the actual appx `InstallLocation`

Testing:
I have tested nano installs and updates and also tested on 2012R2 to ensure current functionality still works as expected. I have tested under the following conditions:
* using a provided `install_msi_url` to force an appx
* using specific versions and also `latest`
* specifying product name and version and channel

Today nano users will have to specify a `install_msi_url` for nano since omnitruck is not serving appx packages. The following urls are publicly accessible:

```
https://s3-us-west-2.amazonaws.com/nano-chef-client/chef-12.14.57.appx
https://s3-us-west-2.amazonaws.com/nano-chef-client/chef-12.14.60.appx
```

One can test this with a Windows Nano Technical Preview 5 vagrant box. I have a packer template that builds these and it is accessible on Atlas via `mwrock/WindowsNano`.

I have tested faking metadata info to ensure that once omnitruck does begin serving appx packages, the code in this PR should work, but in the meantime use a .kitchen.yml like:

```
---
driver:
  name: vagrant

provisioner:
  name: chef_zero
  install_msi_url: https://s3-us-west-2.amazonaws.com/nano-chef-client/chef-12.14.60.appx

platforms:
- name: windows-nano
  driver_config:
    box: mwrock/WindowsNano
```

If testing with Test-Kitchen, use the `nano` branch which includes a very small .net core compatibility adjustment.

### DIFF
--- a/lib/mixlib/install/generator/powershell/scripts/helpers.ps1
+++ b/lib/mixlib/install/generator/powershell/scripts/helpers.ps1
@@ -1,16 +1,22 @@
 function Get-PlatformVersion {
-  switch -regex ((get-wmiobject win32_operatingsystem).version) {
-    '10\.0\.\d+' {$platform_version = '2012r2'}
+  switch -regex ((Get-WMIQuery win32_operatingsystem).version) {
+    '10\.0\.\d+' {$platform_version = '2016'}
     '6\.3\.\d+'  {$platform_version = '2012r2'}
     '6\.2\.\d+'  {$platform_version = '2012'}
     '6\.1\.\d+'  {$platform_version = '2008r2'}
     '6\.0\.\d+'  {$platform_version = '2008'}
   }
+
+  if(Test-Path 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Server\ServerLevels') {
+    $levels = Get-ItemProperty 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Server\ServerLevels'
+    if($levels.NanoServer -eq 1) { $platform_version += 'nano' }
+  }
+
   return $platform_version
 }
 
 function Get-PlatformArchitecture {
-  if ((get-wmiobject win32_operatingsystem).osarchitecture -like '64-bit') {
+  if ((Get-WMIQuery win32_operatingsystem).osarchitecture -like '64-bit') {
     $architecture = 'x86_64'
   } else {
     $architecture = 'i386'
@@ -22,7 +28,8 @@ function New-Uri {
   param ($baseuri, $newuri)
 
   try {
-  new-object System.Uri $baseuri, $newuri
+    $base = new-object System.Uri $baseuri
+    new-object System.Uri $base, $newuri
   }
   catch [System.Management.Automation.MethodInvocationException]{
     Write-Error "$($_.exception.message)"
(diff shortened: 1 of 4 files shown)


lib/mixlib/install/generator/powershell/scripts/helpers.ps1
```diff
@@ -1,5 +1,10 @@
 function Get-PlatformVersion {
-  switch -regex ((get-wmiobject win32_operatingsystem).version) {
+  if(Test-Path 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Server\ServerLevels') {
+    $levels = Get-ItemProperty 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Server\ServerLevels'
+    if($levels.NanoServer -eq 1) { return 'nano' }
```
mwrock: I was thinking about this and planning to change `nano` to `2016nano` based in combo with the actual OS version. The `NanoServer` value above will likely always be `1`. Its just ta flag similar to `ServerCore` but it does appear that releases of nano will be tied to a release of Windows Server. This release is tied to Windows Server 2016.

wrightp:
@mwrock You are the resident expert on this, but I would caution that since we don't have automated testing for the install scripts we need to do our due diligence on testing it. I am in the process of creating an acceptance suite for this repo in PR #133. I'm not sure if it will be ready by the time you need this merged in. Let me know.

lib/mixlib/install/generator/powershell/scripts/helpers.ps1
schisamo: OK so is there really a 3rd dimension to Windows versioning....`server_level` perhaps? So for a Windows Server 2016 Nano edition instance we would have the following:

* platform - `windows`
* platform_version - `x.x.x` (not sure on what `MAJOR.MINOR.PATCH` is)
* server_level - `nano`

Not sure if "level" is the proper term in MS vocab but we should match whatever they use. We should also introduce that to Ohai IMO.

mwrock:
Totally agree @wrightp. This is very much in in flux right now but the tweaks I have made I have been running on nano and non-nano and will be doing more testing as it moves along. A real acceptance suite sounds awesome!

mwrock:
I have this "working" now via the nano branch of Test-Kitchen and a provisioner config of:
```
provisioner:
  name: chef_zero
  install_msi_url: https://s3-us-west-2.amazonaws.com/nano-chef-client/chef-12.14.57.appx
```
This code is very UGLY right now and needs a fair bit of cleanup as well as some more edge case dev and testing. Hoping to allocate work time this week to this to get it into the next DK.

mwrock:
I have removed the WIP and added words to the PR description around caveats and testing details.

mwrock:
cc @chef/engineering-services

scotthain:
@mwrock Are we using any 'only found in new versions of Powershell' features that may cause issues with 2008 and the like? Have you tested on 2012 (no r2) and the 2008s?

sersut:
👍 except @scotthain's comment. 

Can we also squash all commits before merge @mwrock? Thanks 😄

mwrock:
I will make sure to test against 2008 today. All the non-nano platforms should be using the same APIs but it is very good to test in case the actual check fails. And to be honest, I'm not 100% sure this ever workes on 2008 (R1) but can try.

will definitely squash before merge @sersut

mwrock:
Works on 2008R2. Need to setup a "r1" vm.

mwrock:
Works on 2008 after installing powershell v2 which mixlib-install needed before as well.

scotthain:
Thanks @mwrock ! 👍

mwrock:
squashed

PR #137, opened 2016-09-14
Add new s390x mappings

scotthain:
@chef/engineering-services


wrightp:
ugh. brain fart. it does NOT need to be ported.

PR #140, opened 2016-09-19
actually copy the appx install instead of linking

mwrock:
We have noticed nano installs failing on azure nano instances as well as local images built differently from our packer template. There are issues with the ACLs of the appx deployed location that prevent chef from being able to be invoked.

I'd like to believe this is either a bug or an error on our part but documentation here is extremely scarce and I am trying to contact someone at microsoft for some more detailed guidance. If it is by design we may have to eventually move to .zip based installeds but I'm skeptical it will come to that.

In the meantime, this PR simply copies the bits from the appx directory to the omnibus root location instead of just linking to the original location.

### DIFF
--- a/lib/mixlib/install/generator/powershell/scripts/install_project.ps1
+++ b/lib/mixlib/install/generator/powershell/scripts/install_project.ps1
@@ -118,21 +118,22 @@ Function Install-ChefAppx($appx, $project) {
   Add-AppxPackage -Path $appx -ErrorAction Stop
   $package = (Get-AppxPackage -Name $project).InstallLocation
   $installRoot = "$env:SystemDrive/opscode"
-  $link = Join-Path $installRoot $project
-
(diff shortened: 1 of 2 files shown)


mwrock:
cc @chef/engineering-services

scotthain:
Ouch :( 👍 gotta do what you gotta do.

mwrock:
I've emailed some contacts at MS and also have opened this forum post: https://social.technet.microsoft.com/Forums/windowsserver/en-US/1f0c7e48-ba68-40e4-bfe5-1215dba3b465/receiving-1073741790-invoking-ruby-on-azure-tp5-but-works-fine-on-locally-built-tp5-nano-vm?forum=NanoServer

mwrock:
Just to note: Microsoft is aware of this issue and the workaround in this pr is the best approach at this time. A future release will support a better security model for our scenario.

PR #143, opened 2016-09-20
Release 1.2.2 with forced lf line endings

mwrock:
Just for sanity sake I recloned with this branch, built a gem on windows, scp'd it to my linux box, installed it and confirmed no line endings in the files.


schisamo:
:+1:

PR #144, opened 2016-09-21
Add InSpec to the product matrix for availability via chef-ingredient

yzl:
and omnitruck.

cc @arlimus @chris-rock @chef/engineering-services


wrightp:
That unit test kinda sucks because you have to know it's there. We need to make it smarter to check that the unit test list matches exactly (size and items) to what `PRODUCT_MATRIX.products` returns. That will ensure it's not missed in the future.

sersut:
👍 for the change and 👍 on what @wrightp said  😄

PR #145, opened 2016-09-23
Removing extraneous line breaks from Windows install logging.

stuartpreston:
This change simply removes the additional line breaks that appear during installation on the Windows platform, reducing the gaps in the logged output.

Before:
before

After:
after

Signed-off-by: Stuart Preston <stuart@pendrica.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,6 @@ addons:
       - unzip
       - ssh
 
-branches:
-  only:
-  - master
-
 install:
   - echo "skip bundle install"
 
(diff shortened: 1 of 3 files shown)


wrightp:
👍  on this. I'll look into as to why the travis build failed and I'll merge when it's working.

wrightp:
@stuartpreston Could you do me a favor and rebase from master and force push to your branch? I made a change to try and fix the build issue. If it fails again I will merge anyway. Thanks!

PR #147, opened 2016-09-28
add condition to allow omnitruck to accept unstable channel requests

wrightp:
This change will reverted once the omnitruck changes with package router have been deployed to production. This is a stop gap solution so unstable requests that depend on the install scripts (sh/ps1) will work. 

This was discovered during acceptance testing for test-kitchen while trying to install a package from unstable.

@chef/engineering-services @sersut @rhass @mwrock


scotthain:
What's the timeline for the Omnitruck deploy vs the revert of this change?

wrightp:
@scotthain omnitruck deploy will be more towards the end of the year

scotthain:
Spiffy @wrightp 👍

PR #153, opened 2016-10-04
Normalize auto detect platform architectures

scotthain:
@chef/engineering-services 

-from wrightp
With these changes I can GET from local omnitruck server:
`0.0.0.0:8080/stable/chef/metadata?p=el&pv=6&m=i386`

and receive:
```
sha1	f8c10dc7f1fefd4dea60473fb269a889c07d0b0a
sha256	c571b2a1a89ea205c824f23fec14f36b5c209908cd98f1b4a3417e5e68461b2e
url	https://packages.chef.io/files/stable/chef/12.14.89/el/6/chef-12.14.89-1.el6.i386.rpm
version	12.14.89
```

need confirmation that this is the desired result.

### DIFF
--- a/lib/mixlib/install/backend/base.rb
+++ b/lib/mixlib/install/backend/base.rb
@@ -175,6 +175,25 @@ module Mixlib
 
           [platform, platform_version]
         end
+
+        #
+        # Normalizes architecture information that we receive from omnibus.architecture
+        #
+        # @param [String] architecture
+        #
+        # @return String [architecture]
+        def normalize_architecture(architecture)
+          case architecture
+          when "amd64"
+            "x86_64"
+          when "x86", "i86pc", "i686"
+            "i386"
+          when "sun4u", "sun4v"
+            "sparc"
+          else
+            architecture
+          end
+        end
       end
     end
   end
--- a/lib/mixlib/install/backend/package_router.rb
+++ b/lib/mixlib/install/backend/package_router.rb
@@ -42,7 +42,6 @@ module Mixlib
                       else
                         artifacts_for_version(options.product_version)
                       end
-
           windows_artifact_fixup!(artifacts)
         end
 
@@ -225,20 +224,6 @@ Can not find any builds for #{options.product_name} in #{endpoint}.
 
         def omnibus_project
           @omnibus_project ||= PRODUCT_MATRIX.lookup(options.product_name, options.product_version).omnibus_project
-        end
-
-        #
-        # Normalizes architecture information that we receive.
-        #
-        # @param [String] architecture
-        #
-        # @return String [architecture]
-        def normalize_architecture(architecture)
-          if %w{ sun4u sun4v }.include?(architecture)
-            architecture = "sparc"
-          end
-
-          architecture
         end
       end
     end
--- a/spec/mixlib/install/backend/package_router_spec.rb
+++ b/spec/mixlib/install/backend/package_router_spec.rb
@@ -295,4 +295,32 @@ context "Mixlib::Install::Backend::PackageRouter all channels", :vcr do
       expect(package_router.info.url).to match "chef-server-core"
     end
   end
+
+  context "architecture normalization" do
+    let(:channel) { :stable }
+    let(:product_name) { "chef" }
+    let(:product_version) { :latest }
+
+    context "when amd64" do
+      it "returns x86_84" do
+        expect(package_router.normalize_architecture("amd64")).to eq "x86_64"
+      end
+    end
+
+    %w{x86 i86pc i686}.each do |a|
+      context "when #{a}" do
+        it "returns i386" do
+          expect(package_router.normalize_architecture(a)).to eq "i386"
+        end
+      end
+    end
+
+    %w{sun4u sun4v}.each do |a|
+      context "when #{a}" do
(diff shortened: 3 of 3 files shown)


spec/mixlib/install/backend/package_router_spec.rb
```diff
@@ -125,6 +125,23 @@
     end
   end
 
+  context "for a 32 bit version that is mapped to i686" do
+    let(:channel) { :stable }
+    let(:product_name) { "chef" }
+    let(:product_version) { "12.4.1" }
+    let(:platform) { "el" }
+    let(:platform_version) { "5" }
+    let(:architecture) { "i386" }
```
schisamo: This should be `i686` correct?

lib/mixlib/install/backend/package_router.rb
```diff
@@ -236,6 +235,8 @@ def omnibus_project
         def normalize_architecture(architecture)
           if %w{ sun4u sun4v }.include?(architecture)
             architecture = "sparc"
+          elsif %w{ i686 }.include?(architecture)
+            architecture = "i386"
           end
```
schisamo: We need to expand this mapping:

```
if %w{ x86_64 amd64 x64 }.include?(architecture)
  "x86_64"
elsif %w{ i386 x86 i86pc i686 }.include?(architecture)
  "i386"
elsif %w{ sparc sun4u sun4v }.include?(architecture)
  "sparc"
elsif %w{ ppc64le ppc64el }.include?(architecture)
  "ppc64le"
```

I pulled most that logic from:
https://github.com/chef/mixlib-install/blob/v1.2.3/lib/mixlib/install/backend/bintray.rb#L207-L279

I would look that file over and see if you think any other mappings/normalization needs to be done. Please note that the old method was more complicated as we had to infer the architecture from the filename (vs now we have a nice `omnibus.architecture` Artifactory property which contains the architecture value as computed by Omnibus).

lib/mixlib/install/backend/base.rb
```diff
@@ -175,6 +175,25 @@ def normalize_platform(platform, platform_version)
...
         end
+
+        #
+        # Normalizes architecture information that we receive.
+        #
+        # @param [String] architecture
+        #
+        # @return String [architecture]
+        def normalize_architecture(architecture)
+          case architecture
```
wrightp: replaced the if/else logic. It was there for a `fuzzy` search method I added to the Array class. We don't need that anymore.

spec/mixlib/install/backend/package_router_spec.rb
wrightp: If that's what you want then there is a lack of understanding what `normalize_architecture` does. It run post-query. If we request i386 (our normalized data) the omnibus.architecture property may be i686. Then we remap it back to i386.  

If we want mixlib-install to accept different architecture inputs then we have a bigger refactor.

This comes down to requirements. What are we trying to solve?

schisamo: This doesn't test any remapping as is right now. An `omnibus.architecture` of `i686` should be re-mapped to `i386`. This test just says `i386` architecture remaps to `i386`.

Take a look at some example properties from a 32-bit EL6 build.
http://artifactory.chef.co/webapp/#/artifacts/browse/tree/Properties/omnibus-stable-local/com/getchef/chef/12.14.89/el/6/chef-12.14.89-1.el6.i386.rpm

wrightp:
Just more examples we can discuss

Artifact to query as omnibus.architecture value of `i686`
http://artifactory.chef.co/webapp/#/artifacts/browse/tree/Properties/omnibus-stable-local/com/getchef/chef/12.14.89/el/6/chef-12.14.89-1.el6.i386.rpm

```ruby
installer = Mixlib::Install.new(product_name: "chef", version: :latest, channel: :stable, platform: "el", platform_version: "6", architecture: "i386")
=> #<Mixlib::Install:0x007fd46e483470 @options=#<Mixlib::Install::Options:0x007fd46e483448 @options={:product_name=>"chef", :version=>:latest, :channel=>:stable, :platform=>"el", :platform_version=>"6", :architecture=>"i386"}>>
irb(main):002:0> installer.artifact_info
=> #<Mixlib::Install::ArtifactInfo:0x007fd46e100b40 @url="https://packages.chef.io/files/stable/chef/12.14.89/el/6/chef-12.14.89-1.el6.i386.rpm", @md5="c4a788cec39951052878b00f47af28e7", @sha256="c571b2a1a89ea205c824f23fec14f36b5c209908cd98f1b4a3417e5e68461b2e", @sha1="f8c10dc7f1fefd4dea60473fb269a889c07d0b0a", @version="12.14.89", @platform="el", @platform_version="6", @architecture="i386">
```

```ruby
installer = Mixlib::Install.new(product_name: "chef", version: :latest, channel: :stable, platform: "el", platform_version: "6", architecture: "i686")
=> #<Mixlib::Install:0x007fd46c95c818 @options=#<Mixlib::Install::Options:0x007fd46c95c7f0 @options={:product_name=>"chef", :version=>:latest, :channel=>:stable, :platform=>"el", :platform_version=>"6", :architecture=>"i686"}>>
irb(main):004:0> installer.artifact_info
```

PR #156, opened 2016-10-05
mixlib-install exec

wrightp:

manual test. Not going to write automated tests at this time.

@chef/engineering-services 
@tduffield

### DIFF
--- a/bin/mixlib-install
+++ b/bin/mixlib-install
@@ -1,0 +1,14 @@
+#!/usr/bin/env ruby
+
+Signal.trap("INT") { exit 1 }
+
+$:.push File.expand_path("../lib", __FILE__)
+$stdout.sync = true
+
+require "mixlib/install/cli"
+
+begin
+  Mixlib::Install::Cli.start(ARGV)
+rescue StandardError => e
+  abort e.message
+end
--- a/mixlib-install.gemspec
+++ b/mixlib-install.gemspec
@@ -14,20 +14,22 @@ Gem::Specification.new do |spec|
   spec.homepage      = "https://chef.io"
 
   spec.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
-  spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
+  spec.executables   = ["mixlib-install"]
   spec.require_paths = ["lib"]
 
   spec.add_dependency "artifactory"
-  spec.add_dependency "mixlib-versioning"
   spec.add_dependency "mixlib-shellout"
+  spec.add_dependency "mixlib-versioning"
+  spec.add_dependency "thor"
 
+  spec.add_development_dependency "aruba"
   spec.add_development_dependency "bundler"
-  spec.add_development_dependency "rake"
-  spec.add_development_dependency "rspec"
+  spec.add_development_dependency "chefstyle"
   spec.add_development_dependency "pry"
+  spec.add_development_dependency "rake"
   spec.add_development_dependency "rb-readline"
+  spec.add_development_dependency "rspec"
+  spec.add_development_dependency "simplecov"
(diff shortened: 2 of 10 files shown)


bin/mixlib-install
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env ruby
+
+require "mixlib/install"
+require "mixlib/cli"
```
tduffield: Wouldn't we need to add mixlib/cli to deps?

wrightp: done!

bin/mixlib-install
```diff
@@ -0,0 +1,87 @@
...
+  options.merge!(Mixlib::Install.detect_platform)
+end
+
+puts "Querying for product with options: #{options}"
+artifact = Mixlib::Install.new(options).artifact_info
+abort "No results found with options: #{options}" if artifact.nil? || artifact.is_a?(Array)
+
+file = File.join(cli.config[:dir], File.basename(artifact.url))
+puts "Starting download #{artifact.url} to #{file}"
+command = "wget --progress=bar:force #{artifact.url} -O #{file}"
```
schisamo: We can easily do this in pure ruby. Take a look at these two examples:
https://github.com/chef/omnibus/blob/master/lib/omnibus/download_helpers.rb#L29-L100
https://github.com/chef/artifactory-client/blob/master/lib/artifactory/resources/artifact.rb#L459-L497

bin/mixlib-install
```diff
@@ -0,0 +1,87 @@
...
+
+cmd = Mixlib::ShellOut.new(command)
+cmd.live_stream ||= $stdout.tty? ? $stdout : nil
+cmd.run_command
+
+puts "Download complete"
+
+exit 0
```
schisamo: It would be nice to get some unit and functional tests on this CLI. Maybe moving the core logic into subcommand classes will make that testing easier.

The functional test for Omnibus's net fetcher might give you some ideas:
https://github.com/chef/omnibus/blob/master/spec/functional/fetchers/net_fetcher_spec.rb

wrightp:
You're right @schisamo. Caught being lazy. I'll make this more robust and testable.

bin/mixlib-install
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env ruby
+
+require_relative "../lib/mixlib/install/cli"
```
schisamo: We can avoid `require_relative` and more properly set up the environment with the following (this is [what we do in Omnibus](https://github.com/chef/omnibus/blob/master/bin/omnibus)):

```
# Trap interrupts to quit cleanly. See
# https://twitter.com/mitchellh/status/283014103189053442
Signal.trap("INT") { exit 1 }

$:.push File.expand_path("../../lib", __FILE__)
$stdout.sync = true
```

mixlib-install.gemspec
```diff
@@ -14,12 +14,13 @@ Gem::Specification.new do |spec|
...
 
   spec.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
-  spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
+  spec.executables   = ["mixlib-install"]
   spec.require_paths = ["lib"]
 
   spec.add_dependency "artifactory"
   spec.add_dependency "mixlib-versioning"
   spec.add_dependency "mixlib-shellout"
+  spec.add_dependency "thor"
```
schisamo: `thor` is probably a better choice as `test-kitchen` already uses it (and has a dep on `mixlib-install`). This will ensure we don't add an additional `mixlib-cli` (transitive) dep to `test-kitchen`.

bin/mixlib-install
wrightp: yup. not there yet :)

wrightp:
@schisamo Ok, ready for a final review DAWG

PR #158, opened 2016-10-12
Reorder check for Arista EOS above RedHat/Fedora

jerearista:
Signed-off-by: Jere Julian <jere@arista.com>

On some versions of Arista EOS `/etc/redhat-release` also exists.  Moving the check for `/etc/Eos-release` before the check for `/etc/redhat-release` ensures correct OS detection.


jerearista:
CI failed during setup on an openssl command:
```
+openssl aes-256-cbc -K -iv -in ci/es-infrastructure.pem.enc -out es-infrastructure.pem -d
iv undefined
The command "ci/before-script.sh" failed and exited with 1 during .
 ```

wrightp:
I thought I had fixed that issue re: the acceptance tests. Sorry you ran into that. I'm 👍 on this change. I will take another look why it tried to run acceptance from a fork.

PR #163, opened 2016-10-24
add user agent request header support

wrightp:
- all requests will include a default mixlib-install header
- new option to set addition user agent headers
- vcr record new episodes option defaulted to be enabled

Custom headers will look like `mixlib-install/1.1.1 someheader`

@chef/engineering-services

### DIFF
--- a/lib/mixlib/install/backend/package_router.rb
+++ b/lib/mixlib/install/backend/package_router.rb
@@ -147,14 +147,30 @@ Can not find any builds for #{options.product_name} in #{endpoint}.
           uri = URI.parse(endpoint)
           http = Net::HTTP.new(uri.host, uri.port)
           http.use_ssl = (uri.scheme == "https")
-
           full_path = File.join(uri.path, url)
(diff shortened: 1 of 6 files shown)


lib/mixlib/install/backend/package_router.rb
```diff
@@ -147,16 +147,29 @@ def get(url)
...
 
+        def create_http_request(full_path)
+          require "mixlib/install/version"
+
+          request = Net::HTTP::Get.new(full_path)
+          request.add_field("User-Agent", "mixlib-install/#{Mixlib::Install::VERSION}")
+
+          if options.user_agent_headers
+            options.user_agent_headers.each do |agent|
+              request.add_field("User-Agent", agent)
```
schisamo: So [per the RFC](https://tools.ietf.org/html/rfc1945#page-46) we don't want 1..n `User-Agent` request headers but instead want a single `User-Agent` with multiple values delimited with a space.

PR #164, opened 2016-10-26
Add `install_path` value to product details

tduffield:
Get/Set the installation path for any given product. Defaults to
"/opt/[product_name]".

### DIFF
--- a/lib/mixlib/install/product.rb
+++ b/lib/mixlib/install/product.rb
@@ -29,6 +29,8 @@ module Mixlib
         :ctl_command,
         :package_name,
         :product_name,
+        :install_path,
+        :omnibus_project,
       ]
 
       #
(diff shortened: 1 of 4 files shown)


lib/mixlib/install/product.rb
```diff
@@ -218,6 +230,7 @@ def lookup(key, version = :latest)
     omnibus_project "chef-server"
     ctl_command "chef-server-ctl"
     config_file "/etc/opscode/chef-server.rb"
+    install_path "/opt/opscode"
```
schisamo: So `install_path` is different in this instance depending on the version of `chef-server`, we'll want to treat things like `package_name` above (you'll need to update `install_path` to also except a `&block`):

```ruby
install_path do |v|
  if (v < version_for("12.0.0")) && (v > version_for("11.0.0"))
    "/opt/chef-server"
  else
    "/opt/opscode"
  end
end
```

We should also fix `config_file` as it is different depending on the version:

```ruby
config_file do |v|
  if (v < version_for("12.0.0")) && (v > version_for("11.0.0"))
    "/etc/chef-server/chef-server.rb"
  else
    "/etc/opscode/chef-server.rb"
  end
end
```

PR #166, opened 2016-11-03
CLI UX and Testing improvements

schisamo:
CLI improvements include:

* Use `enum` option where appropriate (prints valid options to users)
* Set missing defaults
* Add basic architecture option validation

Testing improvements include:

* Collapse unit and functional specs into single `specs/*` directory
* Use RVM and multiple Ruby versions to drive all testing on Travis
* Update naming of Rake tasks to match idioms used in other projects like [chef/omnibus](https://github.com/chef/omnibus).
* Fix Windows package naming for Inspec package check in `powershell_install_ps1` acceptance suite


lib/mixlib/install/cli.rb
```diff
@@ -32,7 +33,9 @@ def list_versions(product_name, channel)
       option :platform_version,
         aliases: ["-l"]
       option :architecture,
-        aliases: ["-a"]
+        default: "x86_64",
+        aliases: ["-a"],
+        enum: %w{i386 s390x sparc x86_64}
```
scotthain: Can we verify that for solaris, it is expecting `x86_64` and not `i86pc`? I don't recall where that logic stands... (and theoretically it they should both return the same non-sparc package)

schisamo: Nope I think you are correct @scotthain, I'm actually going to add a constant to the Options class to hold the valid archs.

scotthain: Sweet thanks. Otherwise nice work, this is great!

schisamo: Actually it looks like the `i86pc` artifacts have an architecture of `i386`.

scotthain: Technically all of them are "correct", any of the the 3 should return the non-sparc packages.

scotthain: Good call, that makes sense.

PR #170, opened 2016-11-18
Enable platform version compatibility mode for Windows artifacts

### LINKED ISSUE #169: Support for Windows 10?
Went to install chefdk 1.0 using `mixlib-install`, per blog post: https://blog.chef.io/2016/11/17/announcing-release-chefdk-1-0/, but it doesn't seem to be working as expected. I'm using Powershell.

Just dropping this here in case there's some helpful info.
```
C:\> chef -v
Chef Development Kit Version: 0.19.6
chef-client version: 12.15.19
delivery version: master (802e801d920ea6b6d48db735aa7c6e7a6194bea4)
berks version: 5.1.0
kitchen version: 1.13.2

C:\> mixlib-install download chefdk -v 1.0.3
Querying for artifact with options:
{
  "channel": "stable",
  "product_name": "chefdk",
  "product_version": "1.0.3",
  "platform": "windows",
  "platform_version": "2016",
  "architecture": "x86_64"
}
No results found.
```

Yet it is up on https://downloads.chef.io/chef-dk/windows/ at: https://packages.chef.io/stable/windows/2008r2/chefdk-1.0.3-1-x86.msi
 
Versions of other stuff:
```
C:\>  (Get-WmiObject Win32_OperatingSystem).Name
Microsoft Windows 10 Enterprise|C:\WINDOWS|\Device\Harddisk0\Partition2

C:\> uname -a
MSYS_NT-10.0 US90A271W 2.5.0(0.295/5/3) 2016-03-31 18:47 x86_64 Msys

C:\> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      5.1.14393.206
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.14393.206
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

wrightp:
Fixes #169 

Windows artifacts now work properly when platform version compat is enabled. The option is enabled by default for the CLI. There is a backlog item to research how enabling this option by default will affect mixlib-install's API. For now, it seems to be OK with the CLI utility.

Added specs and tested windows versions: 2007 (not valid, but should still work with the code), 2008, 2008r2, 2012, 2012r2, 2016, 2016r2 (doesn't exist, but get the idea)

@schisamo @sersut

### DIFF
--- a/lib/mixlib/install/backend/base.rb
+++ b/lib/mixlib/install/backend/base.rb
@@ -94,18 +94,19 @@ module Mixlib
           artifacts.each do |a|
             return a if a.platform_version == options.platform_version
 
-            # We skip the artifacts produced for windows since their platform
-            # version is always set to 2008r2 which breaks our `to_f` comparison.
-            next if a.platform == "windows"
-
             # Calculate the closest compatible version.
             # For an artifact to be compatible it needs to be smaller than the
             # platform_version specified in options.
             # To find the closest compatible one we keep a max of the compatible
             # artifacts.
+
+            # Remove "r2" from artifacts produced for windows since platform
+            # versions with that ending break `to_f` comparisons.
+            current_platform_version = a.platform_version.chomp("r2").to_f
+
             if closest_compatible_artifact.nil? ||
-                (a.platform_version.to_f > closest_compatible_artifact.platform_version.to_f &&
-                  a.platform_version.to_f < options.platform_version.to_f )
+                (current_platform_version > closest_compatible_artifact.platform_version.to_f &&
+                  current_platform_version < options.platform_version.to_f )
               closest_compatible_artifact = a
             end
           end
--- a/lib/mixlib/install/cli.rb
+++ b/lib/mixlib/install/cli.rb
@@ -36,6 +36,12 @@ module Mixlib
         default: "x86_64",
         aliases: ["-a"],
         enum: Mixlib::Install::Options::SUPPORTED_ARCHITECTURES.map(&:to_s)
+      option :platform_version_compat,
+        desc: "Enable or disable platform version compatibility mode.
+This will match the closest earlier version if the passed version is unavailable.
+If no earlier version is found the earliest version available will be set.",
+        type: :boolean,
+        default: true
       option :url,
         desc: "Print download URL without downloading the file",
         type: :boolean
@@ -48,6 +54,7 @@ module Mixlib
           channel: options[:channel].to_sym,
           product_name: product_name,
           product_version: options[:version],
+          platform_version_compatibility_mode: options[:platform_version_compat],
         }
 
         # Set platform info or auto detect platform
@@ -59,6 +66,7 @@ module Mixlib
           mixlib_install_options[:platform_version] = options[:platform_version]
           mixlib_install_options[:architecture] = options[:architecture]
         else
+
           mixlib_install_options.merge!(Mixlib::Install.detect_platform)
         end
 
--- a/lib/mixlib/install/options.rb
+++ b/lib/mixlib/install/options.rb
@@ -61,6 +61,8 @@ module Mixlib
       def initialize(options)
         @options = options
 
+        map_windows_desktop_versions! if for_windows?
+
         validate!
       end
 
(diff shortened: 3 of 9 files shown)


rhass:
I think the case the user had was with Windows 10, we may want to add a quick check desktop versions just to be safe. Other than that, I am :+1: .

lib/mixlib/install/cli.rb
```diff
@@ -36,6 +36,10 @@ def list_versions(product_name, channel)
         default: "x86_64",
         aliases: ["-a"],
         enum: Mixlib::Install::Options::SUPPORTED_ARCHITECTURES.map(&:to_s)
+      option :platform_version_compat,
+        desc: "Enable or disable platform version compatibility mode when using platform detection",
+        type: :boolean,
+        default: true
```
schisamo: Might be nice to add a short alias here.

lib/mixlib/install/cli.rb
```diff
@@ -36,6 +36,10 @@ def list_versions(product_name, channel)
         default: "x86_64",
         aliases: ["-a"],
         enum: Mixlib::Install::Options::SUPPORTED_ARCHITECTURES.map(&:to_s)
+      option :platform_version_compat,
+        desc: "Enable or disable platform version compatibility mode when using platform detection",
```
schisamo: We might want to mention that this is platform version backward compatibility mode. I don't think that is completely obvious as things are worded now.

lib/mixlib/install/backend/base.rb
```diff
@@ -94,18 +94,19 @@ def filter_artifacts(artifacts)
...
-
             # Calculate the closest compatible version.
             # For an artifact to be compatible it needs to be smaller than the
             # platform_version specified in options.
             # To find the closest compatible one we keep a max of the compatible
             # artifacts.
+
+            # Remove "r2" from artifacts produced for windows since platform
+            # versions with that ending break `to_f` comparisons.
+            current_platform_version = a.platform_version.chomp("r2").to_f
```
schisamo: Would it be better to just do a gsub and remove all non-numeric characters? Something like this should work:

```ruby
=> true
irb(main):003:0> "2008r2".gsub(/[^\d,\.]/, '')
=> "20082"
irb(main):004:0> "2008".gsub(/[^\d,\.]/, '')
=> "2008"
```

wrightp: I don't like this solution because the first example would end up mapping 2008r2 to 2012r2. It should map as accurately as possible. I prefer to explicit chomp.

lib/mixlib/install/cli.rb
wrightp: It's weird to add an alias for a boolean that defaults to true. I believe it will be very rare that `--no-platform-compat` will be used. I don't think it makes sense in this case.

wrightp:
@rhass good call. Done!

lib/mixlib/install/options.rb
```diff
@@ -174,6 +177,23 @@ def validate_user_agent_headers
 
         error
       end
+
+      def map_windows_desktop_versions
```
schisamo: We should add a `!` to the end of this method since `options` is being mutated.

lib/mixlib/install/options.rb
```diff
@@ -174,6 +177,23 @@ def validate_user_agent_headers
...
+                                       "2016"
+                                     when /^6.3/
+                                       "2012r2"
+                                     when /^6.2/
+                                       "2012"
+                                     when /^6.1/
+                                       "2008r2"
+                                     when /^6.0/
+                                       "2008"
+                                     else
```
schisamo: Do we need to include `7` and `8` in here also?

wrightp: I ripped this off from the existing platform detection script? (shrug)

wrightp: Oh I guess they won't map. I don't know if we should or not.

PR #173, opened 2016-12-05
Passing ChefClientFeature in ADDLOCAL parameter

NimishaS:
It's necessary to pass `ADDLOCAL=ChefClientFeature` for complete installation of `chef-client` along with `ChefServiceFeature` and `ChefSchTaskFeature` , otherwise several binaries don't get installed inside `C:/opscode/chef/embedded/bin` folder.


btm:
:+1:

PR #176, opened 2016-12-14
[ET-372]: WIP Rename delivery package to automate - **Do not merge**

PrajaktaPurohit:
**Do not merge**

### DIFF
--- a/PRODUCT_MATRIX.md
+++ b/PRODUCT_MATRIX.md
@@ -10,7 +10,7 @@
 | Chef Server HA Provisioning for AWS | chef-server-ha-provisioning |
 | Chef Development Kit | chefdk |
 | Chef Compliance | compliance |
-| Delivery | delivery |
+| Chef Automate | delivery |
 | Chef Server High Availability addon | ha |
 | Harmony - Omnibus Integration Internal Test Project | harmony |
 | InSpec | inspec |
--- a/lib/mixlib/install/product.rb
+++ b/lib/mixlib/install/product.rb
@@ -193,9 +193,33 @@ PRODUCT_MATRIX = Mixlib::Install::ProductMatrix.new do
 
   product "automate" do
     product_name "Chef Automate"
-    package_name "delivery"
-    ctl_command "delivery-ctl"
-    config_file "/etc/delivery/delivery.rb"
+    package_name do |v|
+    #TODO: Change the version to the promoted release version of automate package
+      v < version_for("8.0.0") ? "delivery" : "automate"
+    end
+    ctl_command do |v|
+    #TODO: Find the version of Chef Automate when the delivery-ctl
+    #      command was aliased to automate-ctl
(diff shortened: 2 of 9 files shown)


schisamo:
@wrightp When you are back tomorrow please give @PrajaktaPurohit some feedback on the direction she is taking on these changes?

lib/mixlib/install/product.rb
```diff
@@ -255,10 +279,34 @@ def lookup(key, version = :latest)
...
+    install_path do |v|
+      if v < version_for("9.0.0")
+      #TODO: Once we change the paths of the config files for automate,
+      #      update the following path to /opt/automate
+        "/opt/delivery"
+      else
+        "/opt/delivery"
+      end
+    end
   end
```
schisamo: We can probably completely kill the `delivery` product entry now. Keeping it around will probably just cause confusion. If a consumer of mixlib-install via something like chef-ingredient uses the old naming they can just update the their use of the `delivery` product key to `automate`.

lib/mixlib/install/product.rb
```diff
@@ -193,9 +193,33 @@ def lookup(key, version = :latest)
 
   product "automate" do
     product_name "Chef Automate"
-    package_name "delivery"
-    ctl_command "delivery-ctl"
-    config_file "/etc/delivery/delivery.rb"
+    package_name do |v|
+    #TODO: Change the version to the promoted release version of automate package
+      v < version_for("8.0.0") ? "delivery" : "automate"
```
PrajaktaPurohit: these versions will be determined once we have an actual release of automate. (the assumption here is that, thatversion will be greator that any of the delivery versions relesed so far)
for now: 8.0.0 is the first version of automate to be released
4.4.4 is the version of delivery where automate-ctl was enabled

schisamo:
Related Omnitruck PR: https://github.com/chef/omnitruck/pull/203

PRODUCT_MATRIX.md
```diff
@@ -10,7 +10,7 @@
 | Chef Server HA Provisioning for AWS | chef-server-ha-provisioning |
 | Chef Development Kit | chefdk |
 | Chef Compliance | compliance |
-| Delivery | delivery |
+| Chef Automate | delivery |
```
wrightp: We should take a hard stance on this. The naming of `delivery` should just be known as `automate` from here on out. Update the key to be `automate.`

lib/mixlib/install/product.rb
```diff
@@ -193,9 +193,33 @@ def lookup(key, version = :latest)
 
   product "automate" do
```
wrightp: Treat the `automate` and `delivery` products as separate entities. No backwards compat logic for which product name they specify. That will remove the copy/paste issue here and logic that is needed here like the ctl name change will be more clear.

wrightp:
Closing PR in favor of #177

PR #179, opened 2017-01-31
[JEX-395] Temporary solution for returning windows desktop artifacts

wrightp:
Since we don't have testers for windows 7, 8, 8.1 or 10 we need to create these as existing artifacts on the fly. We use the same windows msi for all platforms based on architecture.

The desktop version being created dynamically will use the download url for the associated server version. See https://docs.chef.io/platforms.html for details as to why I custom map versions.

Once client SKU testers exist we can remove all this code. It's not very elegant but gets the job done for the time being.

@schisamo @sersut @TrevorBramble @chef/engineering-services 


lib/mixlib/install/backend/base.rb
```diff
@@ -195,6 +221,26 @@ def normalize_architecture(architecture)
...
+        #
+        def map_custom_windows_desktop_versions(desktop_version)
+          unless SUPPORTED_WINDOWS_DESKTOP_VERSIONS.include?(desktop_version)
+            raise UnsupportedVersion, "Unsupported Windows desktop version `#{desktop_version}`. Supported versions: #{SUPPORTED_WINDOWS_DESKTOP_VERSIONS.join(", ")}."
+          end
+
+          server_version = Util.map_windows_desktop_version(desktop_version)
+
+          # Windows desktop 10 officially maps to server 2016.
+          # However, we don't build server 2016 at this time, so we default to 2012r2
```
schisamo: It's actually that we don't _test_ on 2016 at this time so this comment should read:

> However, we don't test on server 2016 at this time, so we default to 2012r2

schisamo: I also updated this PRs description to refer to _testers_ as opposed to _builders_.

spec/unit/mixlib/install/backend/package_router_spec.rb
```diff
@@ -445,4 +445,45 @@
...
+      expect(windows_artifacts.find { |a| a.platform_version == "8.1" }.url).to include "2012r2"
+      expect(windows_artifacts.find { |a| a.platform_version == "10" }.url).to include "2012r2"
+    end
+
+    it "maps architecture to correct filename" do
+      expect(windows_artifacts.find { |a| a.platform_version == "7" && a.architecture == "i386" }.url).to include "-x86"
+      # chefdk doesn't provide 64 bit windows builds
+      expect(windows_artifacts.find { |a| a.platform_version == "7" && a.architecture == "x86_64" }).to be nil
+    end
+  end
```
schisamo: Do we need to add test coverage for `inspec` and `push-jobs-client`?

lib/mixlib/install/backend/base.rb
```diff
@@ -143,6 +152,23 @@ def windows_artifact_fixup!(artifacts)
...
 
+          # Grab windows artifact for each architecture so we don't have to manipulate
+          # the architecture extension in the filename of the url which changes based on product.
+          # Don't want to deal with that!
+          artifact_64 = artifacts.find { |a| a.platform == "windows" && a.architecture == "x86_64" }
+          artifact_32 = artifacts.find { |a| a.platform == "windows" && a.architecture == "i386" }
+
+          SUPPORTED_WINDOWS_DESKTOP_VERSIONS.each do |dv|
+            [artifact_64, artifact_32].each do |artifact|
+              next if artifact.nil?
```
TrevorBramble: You can skip the guard clause if you toss the nils first.
```ruby
[artifact_64, artifact_32].compact.each do |artifact|
```

wrightp:
another hashicorp related error in acceptance. looking into it.

wrightp:
hashie 3.5.0 is the culprit. Works with 3.5.1 which was released shortly after. Whenever travis decides to pick up the latest gem it will work. Don't let this failure stop this merge.

lib/mixlib/install/backend/base.rb
```diff
@@ -195,6 +224,43 @@ def normalize_architecture(architecture)
...
+
+          SUPPORTED_WINDOWS_DESKTOP_VERSIONS.each do |dv|
+            options[:platform_version] = dv
+            options[:url] = base_artifact.url.gsub("\/#{base_artifact.platform_version}\/", "\/#{map_custom_windows_desktop_versions(dv)}\/")
+
+            cloned_artifacts << base_artifact.clone_with(options)
+          end
+
+          cloned_artifacts
+        end
```
TrevorBramble: Fwiw you could always use `Enumerable.each_with_object` for your accumulators, like:
```ruby
        def clone_windows_desktop_artifacts(base_artifact, options = {})
          SUPPORTED_WINDOWS_DESKTOP_VERSIONS.each_with_object([]) do |dv, cloned_artifacts|
            options[:platform_version] = dv
            options[:url] = base_artifact.url.gsub("\/#{base_artifact.platform_version}\/", "\/#{map_custom_windows_desktop_versions(dv)}\/")

            cloned_artifacts << base_artifact.clone_with(options)
          end
        end
```
...though I prefer collect (map) instead as it's simpler and more directly suggests the activity, I think. =^)
```ruby
        def clone_windows_desktop_artifacts(base_artifact, options = {})
          SUPPORTED_WINDOWS_DESKTOP_VERSIONS.collect do |dv|
            options[:platform_version] = dv
            options[:url] = base_artifact.url.gsub("\/#{base_artifact.platform_version}\/", "\/#{map_custom_windows_desktop_versions(dv)}\/")

            base_artifact.clone_with(options)
          end
        end
```

wrightp: I prefer collect as well. Thanks for snapping me out of the each pattern.

PR #182, opened 2017-02-02
Force ASCII encoding for variables.

rhass:
In some situations the strings for various variables are UTF-8 encoded. This causes issues when ruby reads the variables and the strings contain UTF-8 escape sequences and is unable to match strings as needed.


wrightp:
Actually, yeah I bet we could add one. Let me see what I can do.

PR #184, opened 2017-02-23
dynamically resolve platform version compat mode

wrightp:
This PR piggy backs off of platform-options-fix.

There are 3 use cases to consider:
1 - Implementations where platform options and are not configured, and compat mode is defaulted.
2 - Implementations where platform options are set, but compat mode is defaulted.
3 - Implementations where compat mode is set to true.
Use Cases 2 and 3 will be unaffected by this change.
Use Case 1 will be affected in that where results used to return an empty array may now return compatible artifacts, which I would argue is desired behavior. However, this is a change to the existing API.

### DIFF
--- a/lib/mixlib/install/options.rb
+++ b/lib/mixlib/install/options.rb
@@ -66,9 +66,17 @@ module Mixlib
         @options = options
         @errors = []
 
+        resolve_platform_version_compatibility_mode!
+
         map_windows_desktop_versions! if for_windows?
 
         validate!
+      end
+
+      SUPPORTED_OPTIONS.each do |option|
+        define_method option do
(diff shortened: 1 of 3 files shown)


lib/mixlib/install/options.rb
```diff
@@ -118,6 +120,29 @@ def set_platform_info(info)
         validate_options!
       end
 
+      def get_platform_info
```
schisamo: We should drop the `get_*` in the method name.

wrightp:
@schisamo @scotthain After reviewing this change do you guys think it's worth changing at all?

schisamo:
@wrightp I do, we just need to ensure we communicate the potentially breaking change nature of the api update with a major version bump.

PR #185, opened 2017-03-08
Add github_repo and downloads_product_page_url metadata to products

tduffield:
* `github_repo` indicates where the source code can be cloned from. Will be used with Expeditor. 
* `downloads_url` indicates where artifacts can be downloaded (other than via this utility). Will be used with Expeditor.

### DIFF
--- a/lib/mixlib/install/product.rb
+++ b/lib/mixlib/install/product.rb
@@ -20,17 +20,21 @@ require "mixlib/versioning"
 module Mixlib
   class Install
     class Product
-      def initialize(&block)
+      def initialize(key, &block)
+        @product_key = key
         instance_eval(&block)
       end
 
       DSL_PROPERTIES = [
         :config_file,
         :ctl_command,
+        :product_key,
         :package_name,
         :product_name,
         :install_path,
         :omnibus_project,
+        :github_repo,
+        :downloads_product_page_url,
       ]
 
       #
@@ -53,7 +57,7 @@ module Mixlib
               value = instance_variable_get("@#{prop}".to_sym)
               return default_value_for(prop) if value.nil?
 
-              if value.is_a? String
+              if value.is_a?(String) || value.is_a?(Symbol)
                 value
               else
                 value.call(version_for(version))
@@ -77,6 +81,10 @@ module Mixlib
           "/opt/#{package_name}"
         when :omnibus_project
           package_name
+        when :downloads_product_page_url
+          "https://downloads.chef.io/#{product_key}"
+        when :github_repo
+          "chef/#{product_key}"
         else
           nil
         end
@@ -132,7 +140,7 @@ module Mixlib
       # `key` and stores it.
       #
       def product(key, &block)
-        @product_map[key] = Product.new(&block)
+        @product_map[key] = Product.new(key, &block)
       end
 
       #
@@ -162,6 +170,18 @@ module Mixlib
         product.version(version)
         product
       end
+
+      #
+      # Looks up all products that are available on downloads.chef.io
+      #
+      # @return Hash{String => Product}
+      #   :key => product_key
+      #   :value => Mixlib::Install::Product instance
+      def products_available_on_downloads_site
+        @product_map.delete_if do |product_key, product|
+          product.downloads_product_page_url == :not_available
+        end
+      end
     end
   end
 end
@@ -179,16 +199,21 @@ PRODUCT_MATRIX = Mixlib::Install::ProductMatrix.new do
     package_name "opscode-analytics"
     ctl_command "opscode-analytics-ctl"
     config_file "/etc/opscode-analytics/opscode-analytics.rb"
+    downloads_product_page_url :not_available
   end
 
   product "angry-omnibus-toolchain" do
     product_name "Angry Omnibus Toolchain"
     package_name "angry-omnibus-toolchain"
+    github_repo "chef/omnibus-toolchain"
+    downloads_product_page_url :not_available
   end
 
   product "angrychef" do
     product_name "Angry Chef Client"
     package_name "angrychef"
+    github_repo "chef/chef"
+    downloads_product_page_url :not_available
   end
 
   product "automate" do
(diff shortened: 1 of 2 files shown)


lib/mixlib/install/product.rb
```diff
@@ -77,6 +79,10 @@ def default_value_for(prop)
           "/opt/#{package_name}"
         when :omnibus_project
           package_name
+        when :downloads_url
+          "https://downloads.chef.io/#{package_name}"
```
schisamo: This should actually be `"https://downloads.chef.io/#{product_key}"`. This will require we make `product_key` part of `Mixlib::Install::Product`'s `DSL_PROPERTIES`. We should be able to update `Mixlib::Install::Product`'s  initializer to take the key as the first argument:

```ruby
#
# The only DSL method of this class. It creates a Product with given
# `key` and stores it.
#
def product(key, &block)
  @product_map[key] = Product.new(key, &block)
end
```

lib/mixlib/install/product.rb
```diff
@@ -269,6 +286,8 @@ def lookup(key, version = :latest)
       v < version_for("0.7.0") ? "delivery-ctl" : "automate-ctl"
     end
     config_file "/etc/delivery/delivery.rb"
+    github_repo "chef/automate"
+    downloads_url "https://downloads.chef.io/automate"
```
schisamo: This is still a standalone product which is listed separate on downloads.chef.io:
https://downloads.chef.io/compliance

tduffield: Not sure I follow here. This is for delivery, not compliance

schisamo: Yeah looks like I misread the diff....carry on.

tduffield:
@schisamo alright, added product_key

wrightp:
What do you think about being inclusive instead of exclusive regarding the `download_urls`? Rather than having to set `nil` on the packages not available, let's either set 'true' which will result in the default, and also accept a string to the override the url.

lib/mixlib/install/product.rb
```diff
@@ -20,17 +20,21 @@
...
-      def initialize(&block)
+      def initialize(key, &block)
+        @product_key = key
         instance_eval(&block)
       end
 
       DSL_PROPERTIES = [
         :config_file,
         :ctl_command,
+        :product_key,
```
wrightp: Not sure why we are adding `product_key` when `product` maps to the product keys. It looks like we can just use `product` and  override where necessary (downloads_url or github_repo). If I'll take another look, but that is my impression.

tduffield:
@wrightp I'm not sure I agree with that. 

Right now, when defining a product, you either don't do anything (for the default) or you specify `nil` to indicate that there is no downloads url. You also have the ability to specify an alternative string if you want. 

If we do the true/false/string thing, you're now dealing with a property where the boolean is either a boolean or a string, and the output is either a string or a .... what would the output be if the parameter is set to false? would it also be false? Would it be nil?

lib/mixlib/install/product.rb
```diff
@@ -77,6 +81,10 @@ def default_value_for(prop)
           "/opt/#{package_name}"
         when :omnibus_project
           package_name
+        when :downloads_url
+          "https://downloads.chef.io/#{product_key}"
```
schisamo: Actually thinking about this more I don't think we should include a fully qualified URL here but just a boolean flag indicating if the product is displayed on the download site. Maybe just:

```ruby
when :download_site
  true
```

This would allow downloads.chef.io to eventually use this flag in its own logic.

lib/mixlib/install/product.rb
schisamo: We wanted to the key to be available in an instance of `Mixlib::Install::Product`.

wrightp:
Last thoughts and then we can merge this.
1) I would like `downloads_url` to represent what I suppose we deem the "download product page." Like `downloads_product_page_url`. Long yet concise.

2) I always question myself when I want to use symbols in place of POROs. However, I would like to see `downloads_product_page_url` be more clear as to we set set it to nil. Any objections to setting a symbol like `:not_available`?

schisamo:
@wrightp That sounds reasonable, we should probably just go ahead and add a `Mixlib::Install::ProductMatrix.available_on_downloads_page` helper (feel free to come up with a better name) which returns the filtered list of `Mixlib::Install::Product` instances which appear on downloads.chef.io (based on your proposed `downloads_product_page_url` property not being set to `: not_available`).

Why don't you go ahead and update this PR with these final things @wrightp so we can get this merged and shipped for @tduffield.

wrightp:
No prob. I'll finish it up.

wrightp:
New method
```
PRODUCT_MATRIX.products_available_on_downloads_site

=> {"automate"=>"https://downloads.chef.io/automate", "chef"=>"https://downloads.chef.io/chef", "chef-backend"=>"https://downloads.chef.io/chef-backend", "chef-server"=>"https://downloads.chef.io/chef-server", "chefdk"=>"https://downloads.chef.io/chefdk", "compliance"=>"https://downloads.chef.io/compliance", "delivery"=>"https://downloads.chef.io/automate", "inspec"=>"https://downloads.chef.io/inspec", "manage"=>"https://downloads.chef.io/manage", "private-chef"=>"https://downloads.chef.io/private-chef", "push-jobs-client"=>"https://downloads.chef.io/push-jobs-client", "push-jobs-server"=>"https://downloads.chef.io/push-jobs-server", "reporting"=>"https://downloads.chef.io/reporting", "supermarket"=>"https://downloads.chef.io/supermarket"}
```

lib/mixlib/install/product.rb
```diff
@@ -162,6 +170,17 @@ def lookup(key, version = :latest)
         product.version(version)
         product
       end
+
+      def products_available_on_downloads_site
+        product_urls = {}
+
+        products.each do |product|
+          url = lookup(product).downloads_product_page_url
+          product_urls[product] = url if url != :not_available
```
schisamo: Can we just return the complete `Product` instances vs the download product page url only?

Should be as easy as returning a sub-hash:

```ruby
products.delete_if {|product_key, product| product.downloads_product_page_url == :not_available }
```

wrightp: Of course we can.

wrightp:
My initial concern still remains. If we add a new product which is not available on the downloads site and downloads_product_page_url is not set to :not_available this likely will break any implementations (Expeditor) by providing ineligible products for download.

Issue #189, opened 2017-03-14
Return `*.appx` Windows packages for Windows Nano


mwrock:
I noticed that we are now generating chef-client appx packages in artifactory unstable. The chef client pipeline looks to be broken since 2/23 but I'm curious if the pieces are in place once the pipeline clears to drop those packages in current and eventually stable. Looking at the code here, it looks like https://omnitruck.chef.io/current/chef/metadata?v=13&p=windows&pv=2016nano&m=x86_64 would be the URL that would fetch them.

schisamo:
@mwrock What does Ohai return for this particular platform version? Also would it be better to use a marketing name of `nano` as opposed to `2016nano` or do we expect a `2017nano` at some point?

mwrock:
@schisamo nano is tied to windows server 2016. In fact you need t oextract the nano image from a 2016 server install. So conceivably there will be future nanos. Of course its hard to say if they will in fact continue to be called nano.

The ohai data looks identical to a server 2016 installation. There is a registry setting that we check in https://github.com/chef/mixlib-install/blob/64f2b17cb9ef1a4fe214c470bf34bfb0fef07502/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb#L3-L18 to detect nano.

wrightp:
released as part of 3.1

PR #190, opened 2017-03-15
Filter appx packages in search results

scotthain:
Remove appx packages from search results on platforms that do not support appx packages

This is to get around a bug wherein mixlib-install is returning appx packages on 2008r2 (which doesn't support appx). This is not intended to resolve https://github.com/chef/mixlib-install/issues/189

Also I'm pretty sure the added test doesn't... actually do anything. I couldn't make it fail even when changing the expectations :\

@chef/engineering-services

### DIFF
--- a/lib/mixlib/install/backend/base.rb
+++ b/lib/mixlib/install/backend/base.rb
@@ -27,7 +27,7 @@ module Mixlib
         attr_reader :options
 
         SUPPORTED_WINDOWS_DESKTOP_VERSIONS = %w{7 8 8.1 10}
-        SUPPORTED_APPX_PLATFORM_VERSIONS = %w{2016}
+        SUPPORTED_APPX_PLATFORM_VERSIONS = %w{8 8.1 10 2016}
 
         def initialize(options)
           @options = options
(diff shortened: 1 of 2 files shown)


scotthain:
Also worth mentioning - I believe we will want to patch this to a 2.x release if possible.

lib/mixlib/install/backend/base.rb
```diff
@@ -27,6 +27,7 @@ class Base
         attr_reader :options
 
         SUPPORTED_WINDOWS_DESKTOP_VERSIONS = %w{7 8 8.1 10}
+        SUPPORTED_APPX_PLATFORM_VERSIONS = %w{2016}
```
rhass: I think WIndows 8 and Windows 10 also support APPX. @mwrock Can tell us for sure.

rhass:
I don't think Windows 7 is supported by APPX.

schisamo:
@scotthain So for the time being we should just filter out out `*.appx` artifacts completely (for all Windows SKUs).  

A larger, complete, change will come with the fix for #189. Right now we are leaning toward supporting installation of `*.appx` packages on Windows Server 2016 Nano only. All other Windows SKUs will just get `*.msi` files as they do now. Does this sound right to you @mwrock?

I also want @wrightp to weigh in here as he is working on #189 next.

mwrock:
Yes totally agree @schisamo . We should only ship appx to nano and no one else.

scotthain:
@schisamo @mwrock last commit should completely remove all appx packages from return results.

(alsod @tduffield for visibility)

scotthain:
Closing in favor of the immediate solution of removing the offending appx packages from our repo. They shouldn't be there yet anyway.

PR #191, opened 2017-03-16
2016nano appx package support

wrightp:

When 2016nano is configured or auto-detected only appx packages are returned. Otherwise only msi packages are provided for windows platforms.

New acceptance test to validate the install.ps1 script installs appx packages for nano is passing.

This change **can not** be merged with 2.x as this branch already includes the 3.x changes. Apps dependent on these changes will need to update to mixlib-install 3.x.

@schisamo @scotthain @mwrock


lib/mixlib/install/backend/base.rb
```diff
@@ -189,6 +189,11 @@ def windows_artifact_fixup!(artifacts)
             end
           end
 
+          # We currently only support appx packages for 2016nano
+          artifacts.delete_if do |artifact|
+            artifact.url.end_with?(".appx") unless %w{2016nano}.include?(artifact.platform_version)
+          end
+
```
mwrock: do we also need to delete the MSI artifact for nano?

wrightp: I suppose we should if they are built. I added the inverse of this into another block.

mwrock: ah cool

lib/mixlib/install/backend/base.rb
```diff
@@ -189,6 +189,15 @@ def windows_artifact_fixup!(artifacts)
             end
           end
 
+          # We currently only support appx packages for 2016nano
+          artifacts.delete_if do |artifact|
+            artifact.url.end_with?(".appx") unless %w{2016nano}.include?(artifact.platform_version)
```
schisamo: Let's move the list of nano platform versions up in a `SUPPORTED_WINDOWS_NANO_VERSIONS` constant like we have for `SUPPORTED_WINDOWS_DESKTOP_VERSIONS`.

wrightp:
@schisamo I'll test where you noted. I also mistakenly thought the 2016 -> 2012r logic that already was being called for desktop versions. (which it isn't)

lib/mixlib/install/options.rb
```diff
@@ -213,8 +218,15 @@ def validate_platform_options
         end
       end
 
-      def map_windows_desktop_versions!
-        options[:platform_version] = Util.map_windows_desktop_version(platform_version)
+      def map_windows_versions!
+        return unless for_windows?
+
+        # Store original platform version for nano requests
+        if SUPPORTED_WINDOWS_NANO_VERSIONS.include?(options[:platform_version])
```
schisamo: Feels like we should always set `@original_platform_version`, I would remove this guard check.

lib/mixlib/install/backend/base.rb
```diff
@@ -123,14 +121,21 @@ def filter_artifacts(artifacts)
             return closest_compatible_artifact
           end
 
-           # Return an exception if we get to the end of the method
+          # Determine which platform version should be displayed in the exception
+          platform_version = if options.original_platform_version.nil?
+                               options.platform_version
+                             else
+                               options.original_platform_version
+                             end
```
schisamo: Related to https://github.com/chef/mixlib-install/pull/191/files#r106671866 we can skip this code if we just ensure `options.original_platform_version` is always set....we could then just do this below:

```ruby
platform version: #{options.original_platform_version}
```

wrightp: totally was thinking that!

lib/mixlib/install/options.rb
wrightp: aye

.travis.yml
```diff
@@ -14,10 +14,13 @@ addons:
...
   include:
-  - rvm: 2.2.5
-    env: CHEF_ACCEPTANCE=false
-  - rvm: 2.3.0
-    env: CHEF_ACCEPTANCE=true
+  - rvm: 2.2
+  - rvm: 2.3
+    env: SUITE=ubuntu
+  - rvm: 2.3
+    env: SUITE=powershell
```
schisamo: Can we call this suite something like `windows` and the new one `windows-nano`?

acceptance/2016nano/.acceptance/acceptance-cookbook/recipes/provision.rb
```diff
@@ -0,0 +1,41 @@
+run_tf "plan"
+
+run_tf "apply"
+
+ruby_block "install via winrm api" do
+  block do
+    # We do this hackery to ensure we load mixlib-install from source
+    install_command = "install -project angry-omnibus-toolchain -channel current -version 1.1.57"
```
schisamo: Just a heads up `angry-omnibus-toolchain 1.1.57` will eventually be reaped from the `current` channel. We should land a release in the `stable` channel (which is never reaped) before that is an issue.

acceptance/2016nano/inspec/verify.rb
```diff
@@ -0,0 +1,3 @@
+describe file("c:/opscode/angry-omnibus-toolchain") do
+  it { should be_directory }
+end
```
schisamo: Can we add some checks that use [InSpec's command resource](http://inspec.io/docs/reference/resources/command/) to ensure a few of the toolchain's actual tools are installed and are functional. Maybe check for `git`, `ruby` and `gcc`? 

If the`PATH` is configured per:
https://github.com/chef-cookbooks/omnibus/blob/master/recipes/_environment.rb#L24-L28

You can easily just call `--version` on each of those tools...like:
https://github.com/chef-cookbooks/omnibus/blob/master/recipes/_environment.rb#L57-L61

acceptance/2016nano/terraform/application.tf
```diff
@@ -0,0 +1,59 @@
+data "aws_ami" "windows_ami" {
```
schisamo: `windows_server_nano_ami`

acceptance/2016nano/terraform/application.tf
```diff
@@ -0,0 +1,59 @@
...
+    values = ["gp2"]
+  }
+
+  filter {
+    name   = "image-type"
+    values = ["machine"]
+  }
+}
+
+resource "aws_instance" "mixlib_install_ps1" {
```
schisamo: This Terraform resource should be named something different than `mixlib_install_ps1 `...right?

acceptance/2016nano/terraform/application.tf
```diff
@@ -0,0 +1,59 @@
...
+  vpc_security_group_ids = [
+    "sg-96274af3",
+  ]
+
+  tags {
+    # ChefOps's AWS standard tags:
+    X-Dept        = "EngServ"
+    X-Contact     = "pwright"
+    X-Production  = "false"
+    X-Environment = "development"
```
schisamo: ` X-Environment = "acceptance"`

lib/mixlib/install/backend/base.rb
```diff
@@ -146,6 +144,15 @@ def windows_artifact_fixup!(artifacts)
           new_artifacts = [ ]
           native_artifacts = [ ]
 
+          # We only return appx packages when a nano platform version is requested.
+          if options.class::SUPPORTED_WINDOWS_NANO_VERSIONS.include?(options.original_platform_version)
+            return artifacts.find_all { |a| appx_artifact?(a) }
```
schisamo: Let's just add a `#appx_artifact?` method to `Mixlib::Install::ArtifactInfo`.

.travis.yml
wrightp: sure, I'll rename the directory and update this.

acceptance/2016nano/.acceptance/acceptance-cookbook/recipes/provision.rb
wrightp: sure when that happens I'll update the test

acceptance/2016nano/inspec/verify.rb
wrightp: absolultely

acceptance/2016nano/terraform/application.tf
wrightp: these names could be foobar. doesn't really matter. it was a copy/paste remnant. I'll update it though.

acceptance/2016nano/terraform/application.tf
wrightp: I thought `acceptance` was reserved for Delivery acceptance. No?

lib/mixlib/install/backend/base.rb
wrightp: done

acceptance/2016nano/terraform/application.tf
schisamo: @wrightp Nope....it represents the application specified in `X-Application`.

wrightp: ok I'll update the other suites as well

wrightp:
@schisamo I believe I accounted for all your comments. Please review.

schisamo:
This is so awesome for so many reasons...thank you for struggling through how to acceptance test Windows Server Nano using Terraform and InSpec!

Issue #194, opened 2017-04-05
Configurable http_proxy, https_proxy for powershell install script


wrightp:
http_proxy and https_proxy are configurable in test-kitchen for the powershell install script. To maintain parity mixlib-install needs to support configuring the env vars.

Can we safely support adding arbitrary env vars to the install scripts?

wrightp:
I don't particularly like one-off edge case config options. Perhaps it's best if test-kitchen is responsible for setting the env var (`http_proxy`) on the instance via the current provisioner setting.

wrightp:
Another option is to add an `env_vars` option that takes a hash of key/value pairs. The vars could be added to the scripts.

PR #196, opened 2017-04-07
Install script option overrides

wrightp:
The new `install_command_options` option takes a Hash of key/value pairs.

`#install_command`:
- for ps1 this will convert the options into command lines params for the generated install command
- for sh this will convert the options to the generated set of variables

`#install_ps1` and `#install_sh`:
This will take the options and set the `context` Hash which is passed as erb templates attributes.

This feature was added to maintain parity with test-kitchen which supports the provisioner setting `http_proxy`.  However, now we can support overriding any sh variable or ps1 install param.

@chef/engineering-services


wrightp:
#194

Issue #199, opened 2017-04-10
Map sun4v and sun4u to sparc in architecture


scotthain:
Basically doing what this does, only the correct way :D
https://github.com/chef/mixlib-install/pull/198

In essence we should be able to enter `sun4u` or `sun4v` as the architecture parameter and it should map to the `sparc` packages.

@wrightp

wrightp:
merged. This functionality will be available in the next chef-ingredient release

PR #200, opened 2017-04-11
Refactor and expose #normalize_architecture

wrightp:
Related to issue #199 

Also ruby 2.4 and webmock 1.0 is currently broken. I updated webmock to latest (2.0). All is good.

@chef/engineering-services

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ matrix:
     env: SUITE=ubuntu
   - rvm: 2.3
     env: SUITE=windows-server-2012r2
-  - rvm: 2.3
-    env: SUITE=windows-nano
+  # - rvm: 2.3
+  #   env: SUITE=windows-nano
 
 before_script:
   - ci/before-script.sh
--- a/lib/mixlib/install/options.rb
+++ b/lib/mixlib/install/options.rb
@@ -71,7 +71,7 @@ module Mixlib
         @options = options
         @errors = []
 
-        # Store original platform version in cases where we must remap it
+        # Store original options in cases where we must remap
         @original_platform_version = options[:platform_version]
 
         resolve_platform_version_compatibility_mode!
--- a/acceptance/windows-server-2012r2/inspec/verify.rb
+++ b/acceptance/windows-server-2012r2/inspec/verify.rb
@@ -1,3 +1,3 @@
-describe package("Chef Client v12*") do
+describe package("Chef Client v13*") do
   it { should be_installed }
 end
--- a/lib/mixlib/install/backend/base.rb
+++ b/lib/mixlib/install/backend/base.rb
@@ -222,25 +222,6 @@ EOF
           [platform, platform_version]
         end
 
-        #
-        # Normalizes architecture information that we receive from omnibus.architecture
-        #
-        # @param [String] architecture
-        #
-        # @return String [architecture]
-        def normalize_architecture(architecture)
-          case architecture
-          when "amd64"
-            "x86_64"
-          when "i86pc", "i686"
-            "i386"
-          when "sun4u", "sun4v"
-            "sparc"
-          else
-            architecture
-          end
-        end
-
         private
 
         #
--- a/lib/mixlib/install/backend/package_router.rb
+++ b/lib/mixlib/install/backend/package_router.rb
@@ -207,7 +207,7 @@ EOF
                      end
 
           ArtifactInfo.new(
-            architecture:          normalize_architecture(artifact_map["omnibus.architecture"]),
+            architecture:          Util.normalize_architecture(artifact_map["omnibus.architecture"]),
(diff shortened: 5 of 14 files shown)


wrightp:
I may have broken windows :/ On it.

lib/mixlib/install/cli.rb
```diff
@@ -34,8 +34,7 @@ def list_versions(product_name, channel)
         aliases: ["-l"]
       option :architecture,
         default: "x86_64",
-        aliases: ["-a"],
-        enum: Mixlib::Install::Options::SUPPORTED_ARCHITECTURES.map(&:to_s)
+        aliases: ["-a"]
```
schisamo: I feel we should still print a list of supported architectures.

lib/mixlib/install/options.rb
```diff
@@ -25,17 +25,7 @@ class Install
...
-
-      SUPPORTED_ARCHITECTURES = %w{
-        i386
-        powerpc
-        ppc64
-        ppc64le
-        s390x
-        sparc
-        x86_64
-      }
```
schisamo: Does it hurt to keep this list around? It allows us to print nice error messages when validation fails.

wrightp: I don't want to manage this ever.

wrightp: I could always be persuaded. But then why not constrain platforms or platform versions, etc. Let's just let the results do their thing here.

scotthain: I'm kind of on the fence with this... What this really represents is things that map to the `omnibus.architecture` in the metadata, which *should* be a fixed list, so I'm ok with leaving this in there since this is unlikely to change, however the mappings could change.

Does that make sense? I think I'm leaning towards leaving the array since it really shouldn't change. We should be mapping (by exception) to one of these in all cases.

schisamo: This will be a fixed list and thus we should do validation on that fixed liset for an improved UX in mixlib-install and the tools that use it like chef-ingredient.

scotthain: Yeah that's a good distillation @schisamo - I'm +1 on keeping the list

wrightp: cool

wrightp: I believe in asking for forgiveness, not permission. Please forgive me :)

scotthain: No worries @wrightp - good discussion to have! ❤️

wrightp:
Updates on failed acceptance suites.
The 2012r2 suite failed because of a simple Chef 12 check. It has been updated to Chef 13 and now passes.

The nano suite fails because angry-omnibus-toolchain is no longer available in the current repo. I then tried to update the test to use a valid chef 13 appx package. However, there appears to be an issue installing chef 13 appx packages. For now, I am disabling the suite. I will contact Matt Wrock about the install issue. Perhaps there is a specific version that will get us by for a while.

wrightp:
Here's the culprit! https://github.com/chef/chef/issues/6049

.travis.yml
```diff
@@ -19,8 +19,8 @@ matrix:
     env: SUITE=ubuntu
   - rvm: 2.3
     env: SUITE=windows-server-2012r2
-  - rvm: 2.3
-    env: SUITE=windows-nano
+  # - rvm: 2.3
+  #   env: SUITE=windows-nano
```
schisamo: We haven't officially released anything `*.appx` to the `stable` channel. The decision on when to do this should come from @thommay and the community team as they need to be prepared to support it.

lib/mixlib/install/options.rb
```diff
@@ -25,15 +25,20 @@ class Install
...
+        i86pc
         powerpc
         ppc64
         ppc64le
         s390x
         sparc
+        sun4u
+        sun4v
         x86_64
       }
```
schisamo: So I see the issue here now. We don't really want to list these architectures we internally map (`amd64`, `i686`, `i86pc`, `sun4u`, and `sun4v`) as "supported" BUT we want to provide a nice UX and do the right thing. I wonder if the proper place for this automatic mapping to occur is actual in `chef_ingredient` (you could use `Util.normalize_architecture`) as the place we'll need to do this mapping is a Chef recipe as the architecture will come from Ohai. So essentially mixlib-install holds the mapping logic BUT we execute the mapping in `chef_ingredient` before making a call over to mixlib-install.

.travis.yml
wrightp: Understood. I was pulling from current, but the issue was pre-existing. I was hoping to get this test back up and running, so I'm not too concerned about which product or channel. Just need a working appx.

lib/mixlib/install/options.rb
wrightp: Yes!

PR #202, opened 2017-05-02
Add support for download URL overrides

akitada:


akitada:
@wrightp 
Thank you for the review. I've updated the patch to use `install_command_options` as suggested.

akitada:
I've changed the option name to the one introduced in your RFC and squashed commits.

akitada:
@wrightp I've added unit tests for this change. Please note that this PR does not cover Windows.

akitada:
> I'm curious what error is returned when given a bad url.

When it gets 404
```
debian 9 x86_64
downloading https://packages.chef.io/no-such-file
  to file /tmp/install.sh.22379/no-such-file
trying wget...
ERROR 404
Omnitruck artifact does not exist for version latest on platform debian

Either this means:
   - We do not support debian
   - We do not have an artifact for latest

This is often the latter case due to running a prerelease or RC version of chef
or a gem version which was only pushed to rubygems and not omnitruck.

You may be able to set your knife[:bootstrap_version] to the most recent stable
release of Chef to fix this problem (or the most recent stable major version number).

In order to test the version parameter, adventurous users may take the Metadata URL
below and modify the '&v=<number>' parameter until you successfully get a URL that
does not 404 (e.g. via curl or wget).  You should be able to use '&v=11' or '&v=12'
succesfully.

If you cannot fix this problem by setting the bootstrap_version, it probably means
that debian is not supported.

Metadata URL: 
Download URL: https://packages.chef.io/no-such-file
```

Issue #201, opened 2017-04-21
HTTP/HTTPS proxy support?


iancward:
Hello,
I support infrastructure that is behind a firewall which does not allow direct access to the internet. However, we do have HTTP proxies available, and I'm attempting to use Mixlib::Install via pry to make sure it actually works.
Unfortunately, it doesn't not appear to actually work.
```ruby
[18] pry(main)> options = {
[18] pry(main)*   channel: :stable,
[18] pry(main)*   product_name: 'chef',
[18] pry(main)*   http_proxy: 'http://my.proxy.com:8080',
[18] pry(main)*   https_proxy: 'http://my.proxy.com:8080'
[18] pry(main)* }  
=> {:channel=>:stable,
 :product_name=>"chef",
 :http_proxy=>"http://my.proxy.com:8080",
 :https_proxy=>"http://my.proxy.com:8080"}
[20] pry(main)> mi = Mixlib::Install.new(options).detect_platform
=> #<Mixlib::Install:0x0000000272b350
 @options=
  #<Mixlib::Install::Options:0x0000000272b328
   @errors=[],
   @options=
    {:channel=>:stable,
     :product_name=>"chef",
     :http_proxy=>"http://my.proxy.com:8080",
     :https_proxy=>"http://my.proxy.com:8080",
     :platform_version_compatibility_mode=>true,
     :platform=>"el",
     :platform_version=>"7",
     :architecture=>"x86_64"},
   @original_platform_version=nil>>
[21] pry(main)> mi.available_versions
Errno::ETIMEDOUT: Connection timed out - connect(2) for "packages.chef.io" port 443
from /opt/chef/embedded/lib/ruby/2.1.0/net/http.rb:879:in `initialize'
```

My ultimate hope here is that this gem will support http/s proxy and that the chef_client_updater cookbook can be configured to use mixlib::install with a proxy. That way, we can avoid having to use a yum repository mirror of packages.chef.io and have our servers install chef (somewhat) directly via omnitruck.

wrightp:
`http_proxy` is currently only supported for powershell scripts. This was originally added as a feature for Test Kitchen. I'll add this an enhancement.

iancward:
@wrightp if this is something that I can assist with, please let me know. I'd be happy to see this move along, as installing chef client behind a firewall has never really been easy and the value this feature would provide is beyond measure.

wrightp:
[...]

PR #204, opened 2017-05-10
http[s] proxy support

wrightp:
tl;dr add http and https proxy support to mixlib install api.

See Readme updates for how the options interact with different aspects of the application.

Proxy unit and functional tests run against a mock http proxy server.
The windows powershell changes were tested manually using Fiddler.

This change only includes http and https proxy settings. It does not include the full proxy gambit of passing credentials or ftp proxies. This can always be added later.

To add automated acceptance tests a proxy server will need to be provisioned for each proxy test (windows and linux). Let's get this out in the field and think about the best approach for proxy acceptance tests.

- [x] API calls to package router
- [x] install.sh
- [x] install.ps1
- [x] CLI downloads
- [x] acceptance tests (manually for Windows)


wrightp:
#201

wrightp:
@iancward Here the PR for this work. It will take some time as I have to make this work for the installation scripts and download features. Super rough now, but it'll get there.

lib/mixlib/install/backend/package_router.rb
```diff
@@ -149,12 +149,27 @@ def artifacts_for_version(version)
...
+          use_ssl = (uri.scheme == "https")
+
+          response = nil
+          if options.http_proxy
+            proxy_uri = URI.parse(options.http_proxy)
+            begin
+              Net::HTTP.new(uri.host, nil, proxy_uri.host, proxy_uri.port, use_ssl: use_ssl).start do |http|
+                response = http.request(request)
+              end
+            rescue SocketError => e
```
wrightp: I may need to add ConnectionError to this as well...

wrightp:
@smurawski @mwrock - I added you to get your 👀  on the powershell proxy changes.

wrightp:
@chef/engineering-services

lib/mixlib/install.rb
```diff
@@ -64,11 +64,13 @@ def available_versions
     #
     # @return [Array<String>] list of available versions for the given
     # product_name and channel.
-    def self.available_versions(product_name, channel)
+    def self.available_versions(product_name, channel, https_proxy = nil, http_proxy = nil)
```
schisamo: Perhaps it makes sense to make the 3rd parameter an opts hash? This would allow easy adding of more options in the future:

```ruby
#
# List available versions
#
# @param [String] product name
#
# @param [String, Symbol] channel
#
# @param [Hash] opts
# @option opts [String] :https_proxy 
# @option opts [String] :http_proxy 
#
# @return [Array<String>] list of available versions for the given
# product_name and channel.
def self.available_versions(product_name, channel, opts = {})
  Backend.available_versions(
    Mixlib::Install::Options.new(
      product_name: product_name,
      channel: channel.to_sym,
      https_proxy: opts[:https_proxy],
      http_proxy: opts[:http_proxy]
    )
  )
end
```

wrightp:
@iancward Any chance that you to try out this brach and see if the http proxy options work as you would expect?

iancward:
@wrightp wow, I didn't realize how extensive a change like this would be.

The first thing I noticed is that it appears that if you don't specify a URI scheme with the http_proxy (or https_proxy), it defaults to https. I had actually put `http://` at the front of the proxy URL when using an http proxy.

I was able to get the list versions feature to work both via pry/irb and the CLI (which I wish was documented a bit better in the README); however, when I attempted to use the CLI to download the artifact, it failed. It was able to identify what to download (it fetched the latest version just fine), but it looks like the actual download method just uses `Net::HTTP` without any proxy options.

I was able to use `Mixlib::Install.new(options).available_versions` with only the `http_proxy` option specified, but when I tried to use `Mixlib::ShellOut` to use the `Mixlib::Install.new(options).install_command`, it didn't work unless I specified the `https_proxy` option (regardless of whether `http_proxy` was specified).

I have a windows machine accessible to me; however, trying to test this there is proving challenging. (yay  for trying to use ruby, bundler, chef, and git on windows). I'll keep working at it and try to give you a response by early next week.

wrightp:
@iancward Thank you for the response! Incredibly valuable!  I'll try to structure my response as best as I can.

#### URI scheme requirement:
The way the code is currently written the URI scheme (`http://`) would be required. The previous could used all internal URLs so we knew they would all exist, and in fact all be https. That was a missed assumption and something I can fix.

#### CLI download command
I totally missed that part! This bit of code is removed from the rest of the application (and in fact the only place we download anything.) I will fix that as well.

#### CLI docs 
We were hoping the help command would be sufficient, but in reality the tool has grown much in complexity since inception. We could certain add docs for it's capabilities.

#### http vs https proxy combinations
I mimicked much of how the chef client uses proxies. Which may not be the same functionality we really need. There could be an inherent  design issue more than a bug itself. See the section below for how I think I need to fix this moving forward.

### My thoughts on splitting proxy setting responsibilities 
I think something that I may just have to give up on is trying to get the proxy options for the API to also magically work for the scripts. Round peg, square hole kind of thing.  I'm thinking moving forward the proxy options will have to be set differently for API vs install script generation.  The `install_command_options` option is designed exactly for that. For example,
```ruby
install_command_options: { http_proxy: "pserver:80" }
```

I'm worried that this will feel like double-setting a value when in fact they are quite different.  Perhaps good documentation would solve the issue.

iancward:
Is `install_command_options` a `Mixlib::Install` thing or a Chef API thing?
I think the separate `http_proxy` and `https_proxy` options make sense, because curl (and I believe yum as well) will use environment variables for each if they are set; however, I don't remember if curl requires `https_proxy` to be set for for https requests.

iancward:
I upgraded chefdk to 1.0.3 on a windows machine to get the required version of ruby for chefstyle (I was hoping to have Mixlib::Install update it for me), and I was able to run `chef exec bundle exec rake console` and actually get `Mixlib::Install.new(options).available_versions` to work when using the proxy.

However, when I tried to use `Mixlib::ShellOut` to run the `install_command`, I got an error on execution that "The command line is too long". This is on a 64-bit Windows 2008R2 machine. I've received this error running `rake console` from both a cmd prompt and a powershell prompt, and both with and without calling `#detect_platform` prior to instantiating `Mixlib::ShellOut`.

I found a few issues for test kitchen regarding similar errors; however, I'm running the mixlib-install bits directly on the server.
https://github.com/test-kitchen/test-kitchen/issues/811
https://github.com/test-kitchen/test-kitchen/issues/854

The `install_command` powershell command is only 413 lines long, so I'm not quite sure why I'm getting this error. If this was "normal" on Windows, I'd expect to see an issue or something in the chef_client_updater cookbook, but there's no mention of this error there.


I'm also unable to run the `mixlib-install` CLI via bundler (`bundler: not executable: bin\mixlib-install`). I'm not sure if Thor supports Windows; however, there's a CHANGELOG entry from 2008 that claims to have "improved" support for it..

iancward:
Ah I see. There are existing `install_command_options` to override the download URL, and that really only applies to downloading the artifact and not searching for available versions.

wrightp:
This PR is not the correct solution. It seemed straight forward at first, but has become convoluted Before trying to work this further I'm going create a design doc with requirements and go from there.

iancward:
@wrightp thanks for the update. Please let me know if there's any assistance I can provide in terms of requirements, etc.

Issue #188, opened 2017-03-12
Possibly broken architecture detection on Windows


coderanger:
Was doing some testing on 32-bit Windows on EC2 and it was trying to install the x64 build. Didn't do a full debug on it, but the code in helpers.ps1 looks correct at first glance so maybe somewhere else?

If you want to repro try using a Kitchen config something like this:

```yaml
driver:
  name: ec2
  image_id: ami-f6a043e0
  instance_type: m1.medium
```

Given how few people actually use 32-bit Windows this is probably safely very low priority :-)

btm:
I discovered while looking at #205 that Mixlib::Install.detect_platform doesn't validate the output of Mixlib::ShellOut. In that case, it was returning nil values for platform and platform version, which lead mixlib-install to ask for all the platform/platform_version combinations. In the case of #205 this was raising an error related to an unexpected Array.

Looking for `ami-f6a043e0` lead me to [this code in poise-boiler](https://github.com/poise/poise-boiler/blob/master/lib/poise_boiler/helpers/kitchen.rb#L241) that indicates that AMI is a server 2008sp2 image. Basically, mixilb-install doesn't support Server 2008. It doesn't have powershell installed by default, and when you do install it using the server role you get powershell 1.0, which might as well be another language.

coderanger:
@btm Coincidentally 2008 is also the last version of Windows that did 32-bit I think :)

coderanger:
Also Chef does still theoretically support 2008 I think, or at least we list both 2008 and 2008R2 separately in the platform support doc.

tas50:
@coderanger We support 2008R2, but not 2008, which is a bit of a Vista based mess.

coderanger:
We should update https://github.com/chef/chef-rfc/blob/master/rfc021-platform-support-policy.md then :)

coderanger:
Windows 7 does have 32-bit versions I think though, but I'm not sure how to actually test them.

PR #205, opened 2017-05-12
[MSYS:529] for mixlib-install download

piyushawasthi:
1: Added support to windows server 2008 for mixlib-install download.

### DIFF
--- a/lib/mixlib/install/cli.rb
+++ b/lib/mixlib/install/cli.rb
@@ -75,14 +75,15 @@ If no earlier version is found the earliest version available will be set.",
         if options[:url]
           say artifact.url
         else
+          url = artifact.is_a?(Array) ? artifact.map(&:url).first : artifact.url
           FileUtils.mkdir_p options[:directory]
-          file = File.join(options[:directory], File.basename(artifact.url))
+          file = File.join(options[:directory], File.basename(url))
 
           require "json"
           require "net/http"
 
-          say "Starting download #{artifact.url} to #{file}"
-          uri = URI.parse(artifact.url)
+          say "Starting download #{url} to #{file}"
+          uri = URI.parse(url)
           Net::HTTP.start(uri.host) do |http|
             resp = http.get(uri.path)
             open(file, "wb") do |io|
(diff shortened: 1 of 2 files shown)


lib/mixlib/install/cli.rb
```diff
@@ -75,14 +75,15 @@ def download(product_name)
         if options[:url]
           say artifact.url
         else
+          url = artifact.is_a?(Array) ? artifact.map(&:url).first : artifact.url
```
wrightp: The CLI should not be returning a list of ArtifactInfo instances. What problem are you trying to solve here?

piyushawasthi: @wrightp : It's returning an array of hash for windows2008. Hence, I have to verify if the artifact instance is an array or a hash.

wrightp: What I'm indicating is that it should not be, so we need to figure out why.

wrightp:
@piyushawasthi I'm going to close this PR as the issue is not clear to me.
A) What cli command are you executing so the issue can be reproduced?
B) Why are you changing the platform detection script?

piyushawasthi:
@wrightp - The issue related to PowerShell version 2.0 or less which is install in windows server 2008.
A) If you try any command on windows server 2008 or PowerShell version 2.0 that will not work. Like `mixlib-install download chef
`
B) Platform detection script not working on PowerShell version <= 2.0 so we have to add `Set-ExecutionPolicy Bypass -Force`. It will start a PowerShell session that allows for running scripts and keeps the lowered permissions isolated to just the current running process. Also for running remote scripts you need run this script and for PowerShell version <= 2.0 I have to set the -ExecutionPolicy Bypass.

lib/mixlib/install/cli.rb
btm: `Mixlib::Install.detect_platform` => `{:platform=>nil, :platform_version=>nil, :architecture=>nil}` which leads artifact to be an array containing an artifact for 51 platform/platform_version combinations.

This is because Mixlib::Install.detect_platform doesn't validate the results of Mixlib::ShellOut. In one case, powershell.exe doesn't exist and the command fails, but we still put empty values in the hash we return. #188 was apparently against Windows Server 2008sp2, which doesn't have powershell installed by default and when it does you get powershell v1.0 which doesn't take the `-file` option. `type platform_detection.ps1 | powershell -command -` errors too. We should probably use #188 to return an error when Mixlib::ShellOut fails, but @mwrock has reminded me that making these powershell scripts compatible with powershell 1.0 is pretty untenable.

PR #208, opened 2017-05-24
[WIP] replace net/http with http for improved proxy support

wrightp:
Drivers for this change:
- Prepare for proxy enhancements. `http` makes this significantly easier.
- Remove unnecessary exception handling in package router
- Fill testing gaps in package router
- Add HTTPClient for handling all GET requests. It automatically manages SSL and proxies.

**NOTE: proxy support only supported for the API in current state.**


wrightp:
I'm growing very concerned about the vcr file commits...

lib/mixlib/install/options.rb
```diff
@@ -155,6 +160,16 @@ def resolve_platform_version_compatibility_mode!
         end
       end
 
+      def proxy_settings
```
vinyar: Hi, (not expert here). 
I was wondering where will the default come from?
From what I see here: https://github.com/chef/mixlib-install/blob/ee2a0b97ff941606e34d071257ac6f38c48ecb73/lib/mixlib/install/options.rb#L175 they're not inherited from anywhere.

wrightp: Default proxy settings? They will be `nil`. We wouldn't want to default a proxy

vinyar: I meant, would it inherit http_proxy or no_proxy variables from env?

wrightp: mixlib install will not try to read environment variables. They must be explicitly set. It is not intended to work like other data transfer tools (curl) in that manner. Plus, this WIP only includes the proxy settings for using the API. It will get more complex as we add the proxy config for the install scripts.

lamont-granquist:
you're about to get a requirement that mixlib-install runs on ruby 1.9.3 in order to upgrade old chef 11.x clients (all chef 11.x clients were 1.9.3) with the chef_client_updater cookbook -- which would restrict you to (i think) using the v1.x branch of the http gem. 

you probably want to use Net::HTTP::Proxy, which is what chef 11.x used:

https://github.com/chef/chef/blob/11.18.12/lib/chef/http/basic_client.rb#L115-L125

wrightp:
@lamont-granquist acknowledged

PR #210, opened 2017-06-02
Allow mapped architectures (for example i686 or amd64) as valid input…

scotthain:
…, since they will then be mapped to the correct backend arch

@chef/engineering-services

### DIFF
--- a/lib/mixlib/install/options.rb
+++ b/lib/mixlib/install/options.rb
@@ -167,7 +167,7 @@ module Mixlib
       end
 
       def validate_architecture
-        unless architecture.nil? || SUPPORTED_ARCHITECTURES.include?(architecture)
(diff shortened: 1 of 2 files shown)


lib/mixlib/install/options.rb
```diff
@@ -167,7 +167,7 @@ def default_options
       end
 
       def validate_architecture
-        unless architecture.nil? || SUPPORTED_ARCHITECTURES.include?(architecture)
+        unless architecture.nil? || SUPPORTED_ARCHITECTURES.include?(Util.normalize_architecture(architecture))
```
wrightp: Is the input arch coming in from Ohai or from the detect_platform logic... or something else?

scotthain:
After chatting offline - this change likely belongs in chef-ingredient, closing.

PR #213, opened 2017-06-12
Silence the install warning on CI systems.

coderanger:
Given that CI is one of the mentioned exceptions, it would be nice if my build logs didn't have scary things.

I think that syntax should be of the ancients but I defer to a higher authority.

Copying over from https://github.com/chef/omnitruck/pull/110 since that side is dead-ish.


wrightp:
@coderanger just something I noticed in the appveyor docs: 

`APPVEYOR` - `True` if build runs in AppVeyor environment

I'm curious if this is `true` or `True`. It's inconsistent in the appveyor docs: https://www.appveyor.com/docs/environment-variables/

PR #214, opened 2017-06-18
MSYS-529 : Added support for Windows Server 2008 (powershell 1.0) [testing In-progress on powershell v1.0]

piyushawasthi:


coderanger:
Chef no longer supports Windows 2008 (pre r2) so is this aimed at something other than Chef?

piyushawasthi:
@coderanger - This fix is for powershell v1.0. In powershell v1.0 `-File` option doesn't work. Some servers e.g. `Server 2008sp2` have Powershell v1.0 and current code doesn't work on that.Also displaying proper error message if `Mixlib::ShellOut` fails.

coderanger:
Yes, I am aware, but as Chef no longer supports 2008sp2, we should talk about what level of support is being provided. I assume this was requested by a customer which is cool, but since this deviates from the normal code support plan we should lay out some kind of plan. Like will this just be used to install older Chef, or are we expecting use with newer Chef in some (unsupported) way?

wrightp:
@piyushawasthi We will not be accepting this PR for reasons mentioned in this PR and the previous PR that was rejected. You will need to handle this 2008sp2 case via your implementation. It's not not clear *why* effort is being put into this when A) There are no artifacts that would be returned, B) this change artificially returns "2008" as platform version, and C) this code adds a lot of unnecessary noise to the APIs facade interface.

NimishaS:
@btm, can you please add some comments here if these changes are required or not?
@wrightp, we have a task assigned to us (MSYS-529) for adding Server 2008 (powershell 1.0) support in `mixlib-install`.

wrightp:
@NimishaS I've reviewed the ticket and discussed things with @btm. We will need to update the requirements of ticket MSYS-529. Making mixlib install 2008 backwards compatible is not something we will be adding. Since the end goal is to move knife-bootstrap into core you will need to port the existing script: https://github.com/chef/knife-windows/blob/acb113c75fc82f6deb4e25efd94823a15882e873/lib/chef/knife/bootstrap/windows-chef-client-msi.erb. I will work with @btm and @mwrock on updating the requirement.

PR #220, opened 2017-07-03
MSYS-608 Fixes hash check on FIPS enabled Windows servers

piyushawasthi:
Signed-off-by: piyushawasthi <piyush.awasthi@msystechnologies.com>

### DIFF
--- a/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
+++ b/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
@@ -111,23 +111,10 @@ function Test-ProjectPackage {
   [cmdletbinding()]
   param ($Path, $Algorithm = 'SHA256', $Hash)
 
-  if (-not (get-command get-filehash -ErrorAction 'SilentlyContinue')) {
-    function disposable($o){($o -is [IDisposable]) -and (($o | get-member | foreach-object {$_.name}) -contains 'Dispose')}
-    function use($obj, [scriptblock]$sb){try {& $sb} catch [exception]{throw $_} finally {if (disposable $obj) {$obj.Dispose()}} }
-    function Get-FileHash ($Path, $Algorithm) {
-      $Path = (resolve-path $path).providerpath
-      $hash = @{Algorithm = $Algorithm; Path = $Path}
-      use ($c = Get-SHA256Converter) {
-        use ($in = (gi $path).OpenRead()) {
-          $hash.Hash = ([BitConverter]::ToString($c.ComputeHash($in))).Replace("-", "").ToUpper()
-        }
-      }
-      new-object PSObject -Property $hash
-    }
-  }
   Write-Verbose "Testing the $Algorithm hash for $path."
   $ActualHash = (Get-FileHash -Algorithm $Algorithm -Path $Path).Hash.ToLower()
-  Write-Verbose "`tDesired Hash - '$hash'"
+
+  Write-Verbose "`tDesired Hash - '$Hash'"
   Write-Verbose "`tActual Hash  - '$ActualHash'"
   $Valid = $ActualHash -eq $Hash
   if (-not $Valid) {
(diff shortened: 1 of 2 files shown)


lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
```diff
@@ -141,7 +141,7 @@ function Get-SHA256Converter {
     [System.Security.Cryptography.SHA256]::Create()
   }
   else {
-    New-Object -TypeName Security.Cryptography.SHA256Managed
+    New-Object -TypeName Security.Cryptography.SHA256Cng
```
btm: we should add a comment here saying "Using SHA256Cng instead of SHA256Managed because SHA256Managed is not FIPS 140-2 compatible"

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
```diff
@@ -136,15 +131,36 @@ function Test-ProjectPackage {
...
+    }
   }
-  else {
-    New-Object -TypeName Security.Cryptography.SHA256Managed
+  return $hash
+}
+
+function Get-SHA256Converter {
+  if (Is-FIPS) {
+    New-Object -TypeName Security.Cryptography.SHA256Cng
```
btm: Does this not work on some platforms? 2008, right?

It'd be nice to reduce this complexity if we could and have less options/conditionals/paths.

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
```diff
@@ -112,21 +112,16 @@ function Test-ProjectPackage {
...
+    new-object PSObject -Property $hash
   }
   Write-Verbose "Testing the $Algorithm hash for $path."
-  $ActualHash = (Get-FileHash -Algorithm $Algorithm -Path $Path).Hash.ToLower()
+
+  $ActualHash = if (Is-FIPS) {
+    (Get-FileHash $Path $Algorithm).Hash
+  } else {
+    (Get-FileHash -Algorithm $Algorithm -Path $Path).Hash.ToLower()
+  }
```
btm: Both of these are using the Get-FileHash that we define below? Why are we making different calls? What happens if we use `Get-FileHash -Algorithm $Algorithm -Path $Path).Hash.ToLower()` in both situations? I think I renamed it in my testing to make sure it wouldn't collide with another Get-FileHash.

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
piyushawasthi: @btm - yes, That's not working on 2008 for FIPS enabled mode.

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
```diff
@@ -112,21 +112,16 @@ function Test-ProjectPackage {
...
-      $hash = @{Algorithm = $Algorithm; Path = $Path}
-      use ($c = Get-SHA256Converter) {
-        use ($in = (gi $path).OpenRead()) {
-          $hash.Hash = ([BitConverter]::ToString($c.ComputeHash($in))).Replace("-", "").ToUpper()
-        }
-      }
-      new-object PSObject -Property $hash
-    }
+    $hash = Get-FileHash $Path $Algorithm
+    new-object PSObject -Property $hash
```
btm: renaming `$hash` to `$DesiredHash` would make this a little clearer. this felt like cruft until I realized that.

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
```diff
@@ -112,21 +112,12 @@ function Test-ProjectPackage {
...
-      }
-      new-object PSObject -Property $hash
-    }
+    $DesiredHash = Get-FileHash $Path $Algorithm
+    new-object PSObject -Property $DesiredHash
   }
   Write-Verbose "Testing the $Algorithm hash for $path."
   $ActualHash = (Get-FileHash -Algorithm $Algorithm -Path $Path).Hash.ToLower()
+
   Write-Verbose "`tDesired Hash - '$hash'"
```
btm: You'll want to change `$hash` here to `$DesiredHash` as well as two lines down where we compare them.

I wonder about how `$Hash` is taken as a parameter to `Test-ProjectPackage`. Maybe that's cruft? The old code doesn't appear to really support that from a quick glance.

piyushawasthi:
@btm - Changes have done as per your comment also verified on different platform (Windows server 2012, 2016) and working fine as per expectation. Please review the PR.

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
```diff
@@ -112,22 +112,13 @@ function Test-ProjectPackage {
   param ($Path, $Algorithm = 'SHA256', $Hash)
 
   if (-not (get-command get-filehash -ErrorAction 'SilentlyContinue')) {
```
btm: I think you want to remove this if statement now. I don't think we're using it for anything since we're making a Get-FileHash function below.

piyushawasthi: I removed this if statement and local Get-FileHash function because we are using this inside the if block only. Another Get-FileHash is powershell cmdlet.

btm: Don't we have to use our `Get-FileHash` so that we use our `Get-SHA256Converter`? 

The original code added `Get-FileHash` behind that if, so I figured some platforms were missing it. So I wouldn't remove that even if we don't need it.

piyushawasthi: Yes, we need our code otherwise it's failing for FIPS. added back

wrightp:
I don't know much about FIPS. Are there FIPS ec2 images or is this a feature we can provision and enable on any Windows instance? It would be great to add a FIPS acceptance test.

btm:
We had acceptance testers for chef for a while, but they were unstable and removed in https://github.com/chef/chef/commit/0b857d60eb23ac3cb855f17b9ed054cd99eab785.

You can turn any Windows ec2 instance into FIPS enabled by setting a registry key (set HKLM\SYSTEM\CurrentControlSet\Control\Lsa\FipsAlgorithmPolicy\Enabled to 1).

We could duplicate the `windows-server-2012r2` suite and do that via travis if desired.

wrightp:
The acceptance test for this use case could be greatly simplified as we are really just concerned with the install.ps1 downloading and installing the artifact when FIPS is enabled. I'm happy to help.

schisamo:
@btm When @wrightp mentioned acceptance tests he means for this repo not chef.

btm:
Yes, I understood that.

piyushawasthi:
@wrightp - I have squashed commits.

PR #221, opened 2017-07-17
Only override http proxy if a value is supplied. 

rlaveycal:
The chef-client-updater cookbook doesn't work behind a proxy as the `http_proxy` environment variable is overwritten by the Powershell install script being passed a null value.

This simple change just adds a guard condition so that only if a value is passed to the script is it used.

This is related to #201 

This is an Obvious fix.


wrightp:
@rlaveycal I will get this merged once a DCO is signed: https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

PR #226, opened 2017-07-25
[In-progress] MSYS-618 - Performance tunning of Install-Project

piyushawasthi:
Here is  Bench-Mark after these changes.

**Download and install Chef-client With daemon option is service**

VERBOSE: Total Minutes -> 1
VERBOSE: Total Second -> 12

**Download and install Chef-client With daemon option is service on FIPS**

VERBOSE: Total Minutes -> 1
VERBOSE: Total Second -> 19

### DIFF
--- a/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
+++ b/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
@@ -1,7 +1,7 @@
 [Console]::OutputEncoding = New-Object -typename System.Text.ASCIIEncoding
 
 function Get-PlatformVersion {
-  switch -regex ((Get-WMIQuery win32_operatingsystem).version) {
+  switch -regex ((Get-Win32OS).version) {
     '10\.0\.\d+' {$platform_version = '2016'}
     '6\.3\.\d+'  {$platform_version = '2012r2'}
     '6\.2\.\d+'  {$platform_version = '2012'}
@@ -18,12 +18,20 @@ function Get-PlatformVersion {
 }
 
 function Get-PlatformArchitecture {
-  if ((Get-WMIQuery win32_operatingsystem).osarchitecture -like '64-bit') {
+  if ((Get-Win32OS).osarchitecture -like '64-bit') {
     $architecture = 'x86_64'
   } else {
     $architecture = 'i386'
   }
   return $architecture
+}
+
+function Get-Win32OS {
+  if(!$env:win32OS)
+  {
+    $env:win32OS = Get-WMIQuery win32_operatingsystem
+  }
+  $env:win32OS
 }
 
 function New-Uri {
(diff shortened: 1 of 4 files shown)


wrightp:
@piyushawasthi I temporarily switched the merge from master to a custom branch so I could initiate the acceptance tests. I will move it back once they pass. Also, not sure why the travis tests are failing. Seems unrelated to your changes.

wrightp:
The test failures appear to be related to transitive deps. Looking into this. I ran the acceptance tests locally and they passed.

wrightp:
@mwrock Could you please take a moment to look at this PR? Will this have any effect on nano systems?

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
```diff
@@ -110,20 +118,25 @@ function Get-WebContentOnCore {
 function Test-ProjectPackage {
   [cmdletbinding()]
   param ($Path, $Algorithm = 'SHA256', $Hash)
+  if (!$env:Valid_ProjectPackage){
+    Write-Verbose "Testing the $Algorithm hash for $path."
+    $ActualHash = if (Is-FIPS) {
+      (Custom-GetFileHash -Algorithm $Algorithm -Path $Path).Hash.ToLower()
+    } else {
+      (Get-FileHash -Algorithm $Algorithm -Path $Path).Hash.ToLower()
```
mwrock: does the `get-filehash` function still exist?

mwrock:
Have you done perf comparisons looking at time taken with this change and without?

Caching in the environment will only work if you are making multiple calls inside of the same powershell invokation.

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
piyushawasthi: No, Here `Get-FileHash` is powershell utility.  I have renamed  `Get-FileHash` into `Custom-GetFileHash`. I just make less use of that function due to N+ loop between `disposable` and `use` function.

piyushawasthi:
@mwrock - I have caching the environment because there are 2 calls for the same. Also Without these changes `Install-Project -project chef -channel stable -daemon service -Verbose` take time upto 2 minutes (Avg between 1.39 min to 1.53 min) now it will reach around 1 min (between 0.58 to 1.1). 

**Sample snapshot after changes:**
final_reduce

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
mwrock: ahh. right. be aware that cmdlet was introduced in ps v4 and will therefore fail on OS versions prior to 2012R2 unless their ps has been updated.

mwrock:
thanks for the explanation above. That makes sense.

piyushawasthi:
Thanks @mwrock

mwrock:
So just to reiterate the above comment about `Get-FileHash`. That will break on windows 2012 (not R2) or 2008/2008 R2.

wrightp:
@mwrock  Oooh Snap. I need to revert this change in master. @piyushawasthi There will be work to do yes?

piyushawasthi:
@wrightp: I have done with changes but I am verifying on different platform to ensure no issue on any platform. I'll let you know ASAP.

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
piyushawasthi: make sense, Let me verify on ps v2 and ps v3.

PR #228, opened 2017-07-27
Fix for older powershell version

piyushawasthi:
Signed-off-by: piyushawasthi <piyush.awasthi@msystechnologies.com>

### DIFF
--- a/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
+++ b/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
@@ -120,11 +120,7 @@ function Test-ProjectPackage {
   param ($Path, $Algorithm = 'SHA256', $Hash)
   if (!$env:Valid_ProjectPackage){
     Write-Verbose "Testing the $Algorithm hash for $path."
-    $ActualHash = if (Is-FIPS) {
-      (Custom-GetFileHash -Algorithm $Algorithm -Path $Path).Hash.ToLower()
-    } else {
-      (Get-FileHash -Algorithm $Algorithm -Path $Path).Hash.ToLower()
-    }
+    $ActualHash = (Custom-GetFileHash -Algorithm $Algorithm -Path $Path).Hash.ToLower()
 
     Write-Verbose "`tDesired Hash - '$Hash'"
     Write-Verbose "`tActual Hash  - '$ActualHash'"
@@ -150,7 +146,7 @@ function Custom-GetFileHash ($Path, $Algorithm) {
 }
 
 function Get-SHA256Converter {
-  if (Is-FIPS) {
+  if ($(Is-FIPS) -ge 1) {
     New-Object -TypeName Security.Cryptography.SHA256Cng
   } else {
     if($PSVersionTable.PSEdition -eq 'Core') {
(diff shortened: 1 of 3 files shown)


piyushawasthi:
@mwrock, @wrightp - I have manually verified changes on different version on PowerShell (v2, v3 and v4) with and without FIPS but now I am verifying these changes with azure chef extension so I'll let you know when I have done with that (hopefully before EOD).

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
```diff
@@ -120,7 +120,7 @@ function Test-ProjectPackage {
   param ($Path, $Algorithm = 'SHA256', $Hash)
   if (!$env:Valid_ProjectPackage){
     Write-Verbose "Testing the $Algorithm hash for $path."
-    $ActualHash = if (Is-FIPS) {
+    $ActualHash = if ($(Is-FIPS) -ge 1 -or $(Get-PowershellVersion) -le 3) {
```
mwrock: is there a reason not to use `Custom-GetFileHash` for all cases to simplify this?

piyushawasthi: Hi @mwrock - I have verified only with `Custom-GetFileHash` and deploy multiVMs (numberofinstances = 10) using azurerm template and that's working for me but still I am thinking `Get-FileHash` cmdlet is good due to in `Custom-GetFileHash` method many loop between `disposable` and `use` function.

So Please confirm should we only use `Custom-GetFileHash` for all cases?

mwrock: I'd doubt that produces a perf improvement `Get-FileHash` calls `dispose` on its objects as well. See `Get-Command Get-FileHash | fl`. If there is no improvement in perf (always worth casual benchmarking) I'd consolidate to one implementation.

piyushawasthi:
@mwrock - Thanks for quick reply, I have made changes and verified. Please review and let me know in case of any concerns.

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
```diff
@@ -150,7 +146,7 @@ function Custom-GetFileHash ($Path, $Algorithm) {
 }
 
 function Get-SHA256Converter {
-  if (Is-FIPS) {
+  if ($(Is-FIPS) -ge 1) {
     New-Object -TypeName Security.Cryptography.SHA256Cng
```
mwrock: so sorry not to be mentioning this earlier but I'm also wondering if we can consolidate here and just use `SHA256Cng` on fips and non fips. Not sure if we could use it on core too and I wouldn't worry about that. If we could, we wouldn't need a separate code path for fips I don't think.

piyushawasthi: I am verifying on nano server also need to verify if any drawback to use `SHA256Cng` rather than it may be support .Net framework 3.5 or above.

piyushawasthi: Hi @mwrock - I have verified on nano server and neither `New-Object -TypeName Security.Cryptography.SHA256Cng` nor `[System.Security.Cryptography.SHA256]::Create("System.Security.Cryptography.SHA256Cng")` working on that. Also If I just use `SHA256Cng` for all that is failing on `windows server 2008` because by default it comes with `powershell -v2.0` and `.NET Framework version 2.0`. As I already mention above `SHA265Cng` required `.NET Framework version 3.5 and above`. So let me know your thought on it.

mwrock: thanks for checking!

PR #235, opened 2017-08-22
remove item from changelog for 3.4.0

wrightp:
Item removed is part of 3.3.4


wrightp:
@schisamo These labels will essentially result in a single commit only correct?

PR #236, opened 2017-08-23
[WIP] download_url_override and checksum enhancements

wrightp:
**Not tested** Currently building acceptance tests.

When specifying `download_url_override` the `checksum` is now optional. Not specifying checksum will result in downloading the file even if an existing file has been detected on local disk.

Refactored related code to improve code paths given optional checksum setting.

This change was driven by [RFC 091](https://github.com/chef/chef-rfc/blob/master/rfc091-deprecate-kitchen-settings.md). Kitchen does not require a checksum for the `install_msi_url`, and is reasonable to add this feature to the install scripts.  That option is being replaced by `download_url` for bourne and powershell installs.


schisamo:
Looks like a good start!

acceptance/windows-server-2012r2-checksum/terraform/application.tf
```diff
@@ -0,0 +1,94 @@
...
+resource "aws_instance" "mixlib_install_ps1" {
+  count = 1
+
+  ami           = "${data.aws_ami.windows_ami.id}"
+  instance_type = "${var.aws_instance_type}"
+  key_name      = "es-infrastructure"
+
+  associate_public_ip_address = true
+
+  subnet_id         = "subnet-11ac0174" # Planet Releng Public Subnet
```
rhass: Is this the account we want to use for this, I am fine if it is -- I just want to make sure though.

wrightp: For now yes. I'm hoping within the next few weeks to convert the acceptance tests over to using expeditor's workflow. Then we'll switch to the es account.

lib/mixlib/install/generator/bourne/scripts/helpers.sh.erb
```diff
@@ -209,6 +209,8 @@ do_checksum() {
   if exists sha256sum; then
     echo "Comparing checksum with sha256sum..."
     checksum=`sha256sum $1 | awk '{ print $1 }'`
+    echo "$2"
```
wrightp: Remove

wrightp:
Opened a new PR with grouped commits

PR #243, opened 2017-09-26
add #download_artifact method to Mixlib::Install API

wrightp:

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,19 @@ Mixlib::Install.available_versions("chef", "stable")
 # => ["12.13.3", "12.13.7"]
 ```
 
+### Download an artifact
+Download a specific artifact to a configurable location. All platform options (platform, platform_version, architecture) are required  in order to filter a single artifact.
+```ruby
+# detect platform and download to the operating system’s temporary file path
+Mixlib::Install.new(product_name: "chefdk", channel: :stable).detect_platform.download_artifact(Dir.tmpdir)
+# => "/tmp/chefdk-2.3.4-1.deb"
+
+# specify platform options and download to current directory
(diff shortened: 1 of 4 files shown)


lib/mixlib/install.rb
```diff
@@ -83,6 +83,34 @@ def install_command
...
+        raise "Must provide platform options to download a specific artifact"
+      end
+
+      artifact = artifact_info
+
+      FileUtils.mkdir_p directory
+      file = File.join(directory, File.basename(artifact.url))
+
+      uri = URI.parse(artifact.url)
+      Net::HTTP.start(uri.host) do |http|
```
rhass: Won't this have issues with proxies since we are calling Net::HTTP rather than OpenURI?

README.md
```diff
@@ -159,6 +159,19 @@ Mixlib::Install.available_versions("chef", "stable")
 # => ["12.13.3", "12.13.7"]
 ```
 
+### Download an artifact
+Download a specific artifact to a configurable location. All platform options (platform, platform_version, architecture) are required  in order to filter a single artifact.
+```ruby
+# detect platform and download to "/tmp"
+Mixlib::Install.new(product_name: "chefdk", channel: :stable).detect_platform.download_artifact("/tmp")
```
rhass: In the event the example is cargo culted; `Dir.tmpdir`  is a better non-platform specific path rather than `"/tmp"`.

lib/mixlib/install.rb
wrightp: `URI.parse` does the things.

README.md
wrightp: `Dir.tmpdir` is the default value for the parameter in the code.

wrightp: I see what you mean now. I'll update this

PR #249, opened 2018-01-31
Update expeditor config to allow gem builds

scotthain:

I changed the maintainer group since `chef/package-distribution-maintainers` doesn't exist. 

@chef/jex-team


scotthain:
Or should that be @chef/package-router-maintainers ?

Issue #253, opened 2018-04-10
Download fails on Windows when TLS 1.0 is disabled


andy-dufour:
On a Windows server with a .NET version lower than 4.5 if you try and run mixlib-install on a hardened system that only allows TLS1.2 it will fail to download anything as older versions will try TLS 1.0 only/first.

Potential fix here: https://github.com/andy-dufour/mixlib-install/blob/104ed1d071df409adee41246721e4357d39ae8ec/lib/mixlib/install/generator/powershell/scripts/install_project.ps1#L132-L134

andy-dufour:
Fixed by #254

PR #254, opened 2018-04-17
Adding TLS negotiation support for older versions of .NET

andy-dufour:
This should allow older verisons of .NET that are on operating systems forced to negotiate TLS 1.2 to successfully download our software using mixlib-install.

Corrects https://github.com/chef/mixlib-install/issues/253

### DIFF
--- a/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
+++ b/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
@@ -1,4 +1,5 @@
 [Console]::OutputEncoding = New-Object -typename System.Text.ASCIIEncoding
+[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]'Ssl3,Tls,Tls11,Tls12'
 
 function Get-PlatformVersion {
   switch -regex ((Get-Win32OS).version) {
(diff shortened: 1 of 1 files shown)


andy-dufour:
@tas50 @thommay Could I get someone to take a look at this one? I don't think the change is responsible for Travis failing, but I stand to be corrected ;)

andy-dufour:
Pinged the wrong people! @schisamo can you help find someone to review this?

stuartpreston:
@andy-dufour I'm testing this out on a couple of platforms and then should be in a position to approve! Stand by!

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
```diff
@@ -1,4 +1,5 @@
 [Console]::OutputEncoding = New-Object -typename System.Text.ASCIIEncoding
+[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]'Ssl3,Tls,Tls11,Tls12'
```
rhass: I am not sure if older versions of Windows natively support TLS 1.2 or not... but if they do, it should be the only TLS version we use. At the very least we should disable SSLv3, and possibly TLS 1.0.

PR #263, opened 2018-07-25
Test on Ruby head in Travis

tas50:
Rearrange things to match our other projects as well

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
-language: ruby
 sudo: false
+language: ruby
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -12,16 +12,6 @@ addons:
       - unzip
       - ssh
 
-before_install:
```
coderanger: Do we not want this anymore?

tas50: Fixed up now @coderanger

PR #264, opened 2018-07-26
update amazon platform mapping to support Versions 1, 2, and 2 RCs

wrightp:
* Reintroduce el mapping
* Piggyback off of using os-release to determine platform_version
* Fixes https://github.com/chef-cookbooks/chef-ingredient/issues/228


wrightp:
@chef/jex-team @tas50 Please review. Thank you.

wrightp:
I've tested this on amazon linux platforms.

tas50:
I'm really confused by this PR. It goes against the meeting we had. We had agreed to keep the logic from being client side, but this moves the Amazon platform mapping back to the client side where we can't easily update it in the future if something changes on the OS side. Additionally we lose the ability to know how many people are actually requesting packages on Amazon, which is a big question we keep getting asked at a management level. It all looks like RHEL6/7 with this.

PR #268, opened 2018-08-25
Rename suse platform to opensuseleap

tas50:
This aligns the names we use in the install script with ohai. We're already remapping suse to sles in omnitruck. I added a remap for opensuseleap to sles in a PR there. This is a baby step towards a single list of platform names.


tas50:
Updated @wrightp

PR #279, opened 2019-04-15
Add support for armv7l

LocutusOfBorg:
Add also armv7l support
it looks like it makes the testsuite pass on Ubuntu armhf builders...

### DIFF
--- a/lib/mixlib/install/options.rb
+++ b/lib/mixlib/install/options.rb
@@ -29,6 +29,7 @@ module Mixlib
 
       SUPPORTED_ARCHITECTURES = %w{
         aarch64
+        armv7l
         i386
         powerpc
         ppc64


schisamo:
@LocutusOfBorg We don't currently produce any arm builds for Ubuntu so I'm a little confused on what y'all are trying to accomplish with this change.

LocutusOfBorg:
We (Debian and Ubuntu), provide this library in our repository
https://launchpad.net/ubuntu/+source/ruby-mixlib-install
https://packages.qa.debian.org/r/ruby-mixlib-install.html
and we run the testsuite on *each* architecture.
In this case, your package testsuite, stopped working on armhf after some releases (probably because new implemented tests)
http://autopkgtest.ubuntu.com/packages/r/ruby-mixlib-install/disco/armhf

This pull request tries to make it work and pass on this architecture too

schisamo:
@LocutusOfBorg Thanks for the context, sounds like a reasonable change (thank you).

PR #282, opened 2019-05-08
Use a grep command compatible with solaris 10 and older greps

MarkGibbons:
Grep on some solaris servers doesn't support the -q option.

Signed-off-by: markgibbons <mark.gibbons@nordstrom.com>

### Description

Old versions of grep don't support the -q option.  When trying to discover the type of server
a lowest common denominator approach is probably needed.

### Check List

- [x] New functionality includes tests
- [x] All tests pass
- [x] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>
- [x] PR title is a worthy inclusion in the CHANGELOG

### DIFF
--- a/lib/mixlib/install/generator/bourne/scripts/platform_detection.sh
+++ b/lib/mixlib/install/generator/bourne/scripts/platform_detection.sh
@@ -25,7 +25,7 @@
 machine=`uname -m`
 os=`uname -s`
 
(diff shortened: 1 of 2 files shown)


tas50:
Thanks @MarkGibbons

MarkGibbons:
There's a related change in chef core. https://github.com/chef/chef/pull/8507  whenever you get to it.  I need it in chef 14, but don't know how to make that happen.  Really minor problem though.
Thank you, Lamont and Tim

PR #292, opened 2019-09-08
Community distribution fixes

ramereth:
## Description

This provides some fixes which will allow community distributions the ability to easily override various URLs. In addition, this separates out the PRODUCT_MATRIX into it's own file which community distributions could also override as needed.

**NOTE: omnitruck will likely need to be updated to pull in the product matrix properly. I will be providing a PR for that soon. I do not believe dependent gems will run into this issue.**

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:


lib/mixlib/install/dist.rb
```diff
@@ -0,0 +1,29 @@
...
+      # Default github org
+      GITHUB_ORG = "chef".freeze
+      # Bug report URL
+      BUG_URL = "https://github.com/chef/omnitruck/issues/new".freeze
+      # Support ticket URL
+      SUPPORT_URL = "https://www.chef.io/support/tickets".freeze
+      # Resources URL
+      RESOURCES_URL = "https://www.chef.io/support".freeze
+      # MacOS directory name
+      MACOS_DIR = "chef_software".freeze
```
tas50: This probably isn't the best variable name since it's the volume name not the dir name

lib/mixlib/install/dist.rb
```diff
@@ -0,0 +1,29 @@
...
+      # Default project name
+      PROJECT_NAME = "Chef".freeze
+      # Binary repository base endpoint
+      PRODUCT_ENDPOINT = "https://packages.chef.io".freeze
+      # Omnitruck endpoint
+      OMNITRUCK_ENDPOINT = "https://omnitruck.chef.io".freeze
+      # Default product name
+      DEFAULT_PRODUCT = "chef".freeze
+      # Default base product page URL
+      PRODUCT_URL = "https://downloads.chef.io".freeze
```
tas50: shouldn't this be DOWNLOADS_PAGE?

lib/mixlib/install/dist.rb
```diff
@@ -0,0 +1,29 @@
...
+      # Bug report URL
+      BUG_URL = "https://github.com/chef/omnitruck/issues/new".freeze
+      # Support ticket URL
+      SUPPORT_URL = "https://www.chef.io/support/tickets".freeze
+      # Resources URL
+      RESOURCES_URL = "https://www.chef.io/support".freeze
+      # MacOS directory name
+      MACOS_DIR = "chef_software".freeze
+      # Windows directory name
+      WINDOWS_DIR = "opscode".freeze
```
tas50: I'd prefer is this was something like WINDOWS_INSTALL_DIR so it was more clear in the templates below

lib/mixlib/install/dist.rb
ramereth: Would something like ``MACOS_VOLUME`` be acceptable?

tas50: MACOS_VOLUME probably better describes what it is

Issue #339, opened 2020-12-23
Improve mixlib-install to allow other products


Tensibai:
### Describe the Enhancement:
Actually I need two endpoints, one using Chef's mixlib-install and one overriding the mixlib-isntall with cinc project's one.

Improving mixlib-install to take an extra file to allow other products with their api url would avoid that.

### Describe the Need:
I'd wish to have an internal omnitruck endpoint allowing me to install chef or cinc.

### Current Alternative
Monkey patching mixlib-install options like this:

```
module Mixlib
  class Install
    class Options 
      SUPPORTED_PRODUCT_NAMES += %w(cinc auditor cinc-server)

      def initialize(options)
        @options = options
        @errors = []
        
        PRODUCT_MATRIX.product "auditor" do
          product_name "Cinc Auditor"
          package_name "auditor"
          downloads_product_page_url :not_available
        end

        PRODUCT_MATRIX.product "cinc" do
          product_name "Cinc client"
          package_name "cinc"
          downloads_product_page_url :not_available
        end

        PRODUCT_MATRIX.product "cinc" do
          product_name "Cinc client"
          package_name "cinc"
          downloads_product_page_url :not_available
        end
        ENV['PACKAGE_ROUTER_ENDPOINT'] =  ENV.fetch('PACKAGE_ROUTER_ENDPOINT', 'http://packages.cinc.sh') if %w(cinc auditor cinc-server).include? options[:product_name] 
        # Store original options in cases where we must remap
        @original_platform_version = options[:platform_version]

        resolve_platform_version_compatibility_mode!

        map_windows_versions!

        validate!
      
      end
    end
  end
end
```

### Can We Help You Implement This?:
Merge request incoming :)

rjhornsby:
It looks as though support for `cinc` was added to the `chef_client_updater` cookbook, but that cookbook is failing because its dependency on this library does not list cinc in the product matrix. The cookbook correctly installs mixlib-install-3.12.19, but fails with this message when the client version is changed (ie for an upgrade):

```
Generated at 2022-06-17 17:01:54 +0000
Mixlib::Install::Options::InvalidOptions: chef_client_updater[cinc_install_17.10.0] (sev1_cinc_client::install line 1) had an error: Mixlib::Install::Options::InvalidOptions: Unknown product name cinc.
[...]

PR #340, opened 2020-12-23
Allow to extend the supported products of mixlib-install

Tensibai:
Signed-off-by: Tensibai <tensibai@iabis.net>

## Description
Mixlib-install doesn't have a native way to add supported products out of editing the product_matrix.rb file.
This make things complex to address Chef and other distribution install from the same tool (namely for cinc for what I'm insterested about).

This proposal allow to specify an extra file using the Product DSL to add products to the matrix with their own omnitruck url.

## Related Issue
https://github.com/chef/mixlib-install/issues/339

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.

### DIFF
--- a/lib/mixlib/install/options.rb
+++ b/lib/mixlib/install/options.rb
@@ -26,7 +26,7 @@ module Mixlib
     class Options
       class InvalidOptions < ArgumentError; end
 
-      attr_reader :options, :errors, :original_platform_version
+      attr_reader :options, :errors, :original_platform_version, :supported_product_names
 
       SUPPORTED_ARCHITECTURES = %w{
         aarch64
@@ -45,8 +45,6 @@ module Mixlib
         :current,
         :unstable,
       ]
-
-      SUPPORTED_PRODUCT_NAMES = PRODUCT_MATRIX.products
 
       SUPPORTED_SHELL_TYPES = [
         :ps1,
@@ -76,6 +74,15 @@ module Mixlib
 
         # Store original options in cases where we must remap
         @original_platform_version = options[:platform_version]
+
+        # Eval extra products definition
+        extra_products = ENV.fetch("EXTRA_PRODUCTS_FILE", nil)
+        unless extra_products.nil?
+          PRODUCT_MATRIX.instance_eval(::File.read(extra_products), extra_products)
+        end
+
+        # Store supported product names
+        @supported_product_names = PRODUCT_MATRIX.products
 
         resolve_platform_version_compatibility_mode!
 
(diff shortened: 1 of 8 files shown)


lib/mixlib/install/options.rb
```diff
@@ -77,6 +75,15 @@ def initialize(options)
         # Store original options in cases where we must remap
         @original_platform_version = options[:platform_version]
 
+        # Eval extra products definition
+        extra_products = ENV.fetch("EXTRA_PRODUCTS_FILE", nil)
+        unless extra_products.nil?
+          PRODUCT_MATRIX.instance_eval(::File.read(extra_products), extra_products)
+        end
```
ramereth, line 82: I'm not sure how we would use this in practice on the Cinc side. As it stands now, it would require that environment variable to be set so that it would work. I wonder if there is a better way to handle this so that it works seamlessly in practice.

However, with the inclusion of the new `api_url` parameter, we could adjust our version of this to just add ours using our URLs I think.

Tensibai: On cinc client we could just add the file for cinc products and define the env variable to this file path in the dist.rb file.

Using an environment variable allow to use omnitruck API with extra products by setting the var too.

With this we may update the mixlib-install in any chef-client with recent enough ruby (2.5 because of chef-utils ?) and add the environment variable before calling the omnitruck-updater cookbook to migrate from chef to cinc or cinc to chef (as actual cinc clients have no idea of chef products).

ramereth: I'd like to see how this works on our end before we commit to it. Can you do that please?

Tensibai: So to summarize our talk on slack: That's not really easy to do from this PR as using it means cutting off on our overriding of mixlib-install and thus removing the tweaks we make on Gemfile to use cinc's version of mixlib install.

This change would allow using chef's mixlib-install gem and provide an extra file along with an environment variable in workstation commands to extend the amount of products handled.

My main interest in that is stopping to monkey patch either chef or cinc mixlib-install gem to address both products in the evolution I'm doing on omnitruck software for internal use.
It's a fork as Chef as been super clear about not handling alternative distributions on their endpoints, not even with a redirect, so no PR to come to upstream omnitruck to handle that, but as most things are handled by mixlib-install, that's be cool to avoid ugly patching and have a native way to do it.

Tensibai:
@chef/chef-foundation-owners Something I should do on this ?

marcparadise:
Pending conversation between @Tensibai and @ramereth to figure out best way forward.

jaymzh:
Update for followers - @ramereth is gonna chat with @Tensibai this week so we can get this sorted.

lib/mixlib/install/options.rb
ramereth: OK @Tensibai provided [this example](https://gitlab.com/cinc-project/upstream/chef/-/merge_requests/17/diffs) as a use case and it makes sense to me. So this has my approval now.

Tensibai: Exemple implementation on cinc side: https://gitlab.com/cinc-project/upstream/chef/-/merge_requests/17

jaymzh (dismissed):
OK, we have agreement on the approach here, but @Tensibai you need to rebase here...

Tensibai:
@jaymzh Rebased

jaymzh (approved):
I like this. @marcparadise - can you also double-check?

marcparadise (approved):

Issue #330, opened 2020-09-02
Omnitruck Windows install.ps1 script breaks when no_proxy contains *.domain


rlaveycal:
# Version:

Omnitruck

# Environment:

Windows

# Scenario:

Running the [omnitruck](https://omnitruck.chef.io/install.ps1) script fails if the `$env:no_proxy` environment variable contains wildcard domain entries like `*.chef.io`.

# Steps to Reproduce:

The cause is the `Get-WebContentOnFullNet` helper function. It doesn't sanitise the setting of `$proxy.BypassList` which must be an array of regular expressions.

```powershell
$proxy = New-Object -TypeName System.Net.WebProxy

$proxy.BypassList = @( "*.chef.io" ) # will fail

$proxy.BypassList = @( "\*.chef.io" ) # will work

$proxy.BypassList = @( " *.chef.io" ) # also works
```

# Expected Result:

Chef is installed

# Actual Result:

Exception thrown - `Exception setting "BypassList": "parsing "*.chef.io" - Quantifier {x,y} following nothing."`

PR #341, opened 2020-12-29
Fix issue with bypass list containing * entries

erickuiper:
## Description
If the no_proxy list contains an entry with an wildcard it throws the following error:
```
Exception setting "BypassList": "parsing "*.chef.io" - Quantifier {x,y} following nothing."
At C:\opscode\chef_upgrade.ps1:121 char:3
+      $proxy.BypassList = $byPassList
+      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], SetValueInvocationException
    + FullyQualifiedErrorId : ExceptionWhenSetting
```
## Related Issue
[Issue 330](https://github.com/chef/mixlib-install/issues/330)

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.

### DIFF
--- a/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
+++ b/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
@@ -82,10 +82,18 @@ function Get-WebContentOnFullNet {
   
 
   if($bypassList -ne $null){
-     
-	 $bypassList = $bypassList.split(",")
-	 $proxy.BypassList = $byPassList
-  }
+  	$bypassList = $bypassList.split(",")
+    ForEach ($bypassListItem in $bypassList){
+        try {
+          [regex]$bypassListItem | Out-Null
+        }
+        catch {
+          Write-Warning "Wildcard entry $bypassListItem will be ignored, please use regex"
+          Continue
+        }
+				$proxy.BypassArrayList.Add($bypassListItem)
+			}
+    }
   
   $wc.Proxy = $proxy
 
--- a/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
+++ b/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
@@ -78,7 +78,7 @@ function Get-WebContentOnFullNet {
   $wc = new-object System.Net.WebClient
   $wc.Headers.Add("user-agent", "<%= user_agent_string %>")
   $proxy.Address = $env:http_proxy
-  $bypassList = $proxy.BypassList = $env:no_proxy.replace('.', '\.').replace('..', '\..+\.').replace('*', '.+').split(',').Trim()
+  $proxy.BypassList = $env:no_proxy.replace('.', '\.').replace('..', '\..+\.').replace('*', '.+').split(',').Trim()
   $wc.Proxy = $proxy
 
   if ([string]::IsNullOrEmpty($filepath)) {
--- a/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
+++ b/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
@@ -78,23 +78,7 @@ function Get-WebContentOnFullNet {
   $wc = new-object System.Net.WebClient
   $wc.Headers.Add("user-agent", "<%= user_agent_string %>")
   $proxy.Address = $env:http_proxy
-  $bypassList = $env:no_proxy
-  
-
-  if($bypassList -ne $null){
-  	$bypassList = $bypassList.split(",")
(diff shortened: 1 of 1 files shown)


tas50:
@ekuiper-sbp Any chance you can squash this down to a single commit?

ekuiper-sbp:
The commits have been squashed @tas50

mwrock:
I'm sure this gets rid of the error. My only concern is that according to the `BypassList` [docs](https://docs.microsoft.com/en-us/dotnet/api/system.net.webproxy.bypasslist?view=net-5.0), the array should contain valid regex expressions which is why `*.chef.io` would fail. Using `[a-z]+\.chef\.io$` should accomplish the same without error. If anyone is currently using regex expresions, this may break them.

I'm also curious if using `BypassArrayList` with a wild card actually effectively bypasses all the expected domains.

ekuiper-sbp:
Did some testing with the System.Net.WebProxy object and the results were that both RegEx and wildcard entries are accepted, but only RegEx creates an actual exception. 

Accepting this change will result in fixing the bug not accepting wildcards, but will not result in effective implementation of wildcard proxy exceptions. 

In our case it would resolve the issue as we do not require these exceptions in the context of this script. As a workaround I'd have to write a wrapper to re-write the proxy exceptions. Changing the no_proxy would break other solutions like the netsh winhttp proxy config as this only accept wildcard entries.

mwrock (changes requested):
I think I'm 👎 on this as it stands now. It feels like the wrong approach to suppress an error when submitting a proxy exception that will fail. Alternatively perhaps you could catch the error and then fallback to this approach emitting a warning that there are proxy entries that will not be honored.

ekuiper-sbp:
Added a warning when wildcard entries are used, this will not trigger the try catch and thus still continue to execute the script.

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
```diff
@@ -82,10 +82,14 @@ function Get-WebContentOnFullNet {
...
 
   if($bypassList -ne $null){
-     
-	 $bypassList = $bypassList.split(",")
-	 $proxy.BypassList = $byPassList
-  }
+  	$bypassList = $bypassList.split(",")
+    ForEach ($bypassListItem in $bypassList){
+        if ($bypassListItem -match '\*\..*'){
+          Write-Warning "Wildcard entry $bypassListItem will be ignored, please use regex"
```
mwrock, line 88: In some cases a `*` followed by `.` is valid. perhaps just test if the item validates to a propper regex string:

```
try {
  "test" -match $bypassListItem
} catch {
  Write-Warning "$bypassListItem will be ignored, please use a valid regular expression"
}

Comment (ekuiper-sbp) [reply lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb:91]: Have created and tested a proper test using [regex] and added a continue so that the invalid entry is skipped.

File: lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb [hunk 2]
```diff
@@ -82,10 +82,18 @@ function Get-WebContentOnFullNet {
...
-	 $proxy.BypassList = $byPassList
-  }
+  	$bypassList = $bypassList.split(",")
+    ForEach ($bypassListItem in $bypassList){
+        try {
+          [regex]$bypassListItem | Out-Null
+        }
+        catch {
+          Write-Warning "Wildcard entry $bypassListItem will be ignored, please use regex"
+          Continue
```
mwrock, line 92: You could avoid the continue by moving the `$proxy.BypassArrayList.Add($bypassListItem)` into the `try` after the test, but not a blocker.

lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
```diff
@@ -82,10 +82,18 @@ function Get-WebContentOnFullNet {
...
+        }
+				$proxy.BypassArrayList.Add($bypassListItem)
+			}
+    }
```
mwrock, line 96: just need to clean up indentation here.

tpowell-progress:
Per the discussion and the interface on [BypassList](https://docs.microsoft.com/en-us/dotnet/api/system.net.webproxy.bypasslist?view=net-5.0) we are going to reject this PR... conversion of wildcards to regex could potentially add to the confusion.

You are welcome to open a PR to present a cleaner error to the user.

PR #344, opened 2021-01-21
Make Chef omnitruck endpoint configurable

bdwyertech:
## Description
Allow Chef Omnitruck endpoint to be configurable.  Went down the rabbit hole pretty deep trying to pass this from test-kitchen but can't find a way.

## Related Issue
https://github.com/test-kitchen/test-kitchen/issues/1698

## Types of changes

## Checklist:

### DIFF
--- a/lib/mixlib/install/dist.rb
+++ b/lib/mixlib/install/dist.rb
@@ -7,7 +7,7 @@ module Mixlib
       # Binary repository base endpoint
       PRODUCT_ENDPOINT = "https://packages.chef.io".freeze
       # Omnitruck endpoint
-      OMNITRUCK_ENDPOINT = "https://omnitruck.chef.io".freeze
+      OMNITRUCK_ENDPOINT = (ENV["CHEF_OMNITRUCK_ENDPOINT"] || "https://omnitruck.chef.io").freeze
       # Default product name
       DEFAULT_PRODUCT = "chef".freeze
       # Default download page URL


lamont-granquist:
I think I'm :-1: on the env var here.  This should be a parameter that callers can specify to mixlib-install and that should be exposed as a public API and TK needs to be modified to support it.

bdwyertech:
> I think I'm 👎 on the env var here. This should be a parameter that callers can specify to mixlib-install and that should be exposed as a public API and TK needs to be modified to support it.

I'm with you, it's certainly ugly. This PR is really just to demonstrate the issue.  There is another constant elsewhere which has an undocumented env var override so I used the same method.

This must be resolved before the deprecated logic in Test-Kitchen can be removed, otherwise my org is borked (were a large enterprise customer).

Tensibai:
Alternative approach in https://github.com/chef/mixlib-install/pull/340 (would allow to override the endpoint for specific products too)

jaymzh:
@bdwyertech - I agree with the others we want a more robust solution here. Looks like #340 got dropped, so we've picked that back up, hopefully we can get something like that in ASAP. It seems like it was designed for Cinc's case, but can you confirm it'll solve the TK case?

bdwyertech:
Ò

> @bdwyertech - I agree with the others we want a more robust solution here. Looks like #340 got dropped, so we've picked that back up, hopefully we can get something like that in ASAP. It seems like it was designed for Cinc's case, but can you confirm it'll solve the TK case?

If #340 was merged, it looks like it would resolve this issue as well.

tpowell-progress:
Closing in favor of #340

PR #358, opened 2021-04-16
Strip query parameters before parsing URI path

gscho:
## Description

The previously submitted solution breaks when there is a `/` in an auth token. This time we first remove everything after the first encountered `?` and then parse the last item in the URI's path.

## Related Issue
N/A

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:


jeremy-clerc:
Hello,
Could you tell me how you tested this new sed / the url you are trying to strip ? So I can try to fix it and make it work with your use case while not breaking kitchen cf #368.

Thanks

jeremy-clerc:
Ok I think it is the ` that is in the wrong place

PR #373, opened 2021-12-17
Update support to include amazon linux 2022

jeremiahsnapp:
Signed-off-by: Jeremiah Snapp <jeremiah@chef.io>


tas50 (approved):

tas50:
You might want to try this on a box first. I ran into some issues with it when I tried.

PR #380, opened 2022-08-05
Match on 64 for architecture because of different languages

tpowell-progress:
Signed-off-by: Thomas Powell <powell@progress.com>

## Description
Installs on non-English Windows may fail to match on "64-bit" due to localized arch names, but especially non-Latin alphabet versions, e.g., Chinese 64-bit obarchitecture is returned as "64 位"

Match on "64" only for confirming 64-bit install. 32-bit should be the exceptional case at this point, anyway.

## Related Issue
INFC-220 

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] I have read the **CONTRIBUTING** document.
- [X] I have run the pre-merge tests locally and they pass.
- [X] All new and existing tests passed.
- [X] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

### DIFF
--- a/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
+++ b/lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
@@ -20,7 +20,7 @@ function Get-PlatformVersion {
 }
 
 function Get-PlatformArchitecture {
-  if ((Get-Win32OS).osarchitecture -like '64-bit') {
+  if ((Get-Win32OS).osarchitecture -match '64') {
     $architecture = 'x86_64'
   } else {
     $architecture = 'i386'


lib/mixlib/install/generator/powershell/scripts/helpers.ps1.erb
```diff
@@ -20,7 +20,7 @@ function Get-PlatformVersion {
 }
 
 function Get-PlatformArchitecture {
-  if ((Get-Win32OS).osarchitecture -like '64-bit') {
+  if ((Get-Win32OS).osarchitecture -match '64') {
```
marcparadise, line 23: ARM64 is not a concern, right?

marcparadise: We would need further modifications to this script to support arm64 in any case, and this change will not make that situation any worse.

marcparadise (approved):

PR #381, opened 2022-09-16
Remove support for versions of ruby prior to 2.6

marcparadise:
Ruby 2.5 and earlier are all EOL and are no longer receiving security updates.

Signed-off-by: Marc A. Paradise <marc.paradise@gmail.com>


vkarve-chef (approved):

jaymzh (approved):

marcparadise:
Note that windows build was failing prior to this PR, and is not addressed here.

marcparadise:
Will update this further - even though we can't test these older versions, there may still be cases where they get pulled in and will work (like chef-install cookbook for upgrades).   After update, it will leave in 'soft' support (conditional gems by version) but remove the unsupported ruby versions.

jaymzh:
@marcparadise ping? :)

jaymzh:
This relates to: https://github.com/chef/mixlib-install/issues/400

@marcparadise - I'm unclear on the status of this PR - I think we can merge, but it's been 2 years, so we should at least rebase.

mixlib-install.gemspec
```diff
@@ -16,7 +16,7 @@ Gem::Specification.new do |spec|
   spec.files         = %w{LICENSE Gemfile Rakefile} + Dir.glob("*.gemspec") + Dir.glob("{bin,lib,support}/**/*")
   spec.executables   = ["mixlib-install"]
   spec.require_paths = ["lib"]
-
+  spec.required_ruby_version = ">= 2.6.0"
```
jaymzh, line 19: I  _think_ we decided not to force newer versions so people can use this on like an old Mac to get the latest chef... we wanted to go "soft" min requirement.

Gemfile
```diff
@@ -2,22 +2,14 @@ source "https://rubygems.org"
...
 
-gem "chef-utils", "= 16.6.14" if RUBY_VERSION < "2.6.0"
-
 group :test do
   gem "rake"
   gem "rspec"
   gem "webrick"
   gem "webmock", "~> 3.4"
   gem "contracts", "~> 0.16.0" # this entry can go away when ruby < 3 support is gone
-  if RUBY_VERSION < "2.6.0"
```
jaymzh, line 13: probably wanna leave the conditional here....

Gemfile
```diff
@@ -2,22 +2,14 @@ source "https://rubygems.org"
 
 gemspec
 
-gem "chef-utils", "= 16.6.14" if RUBY_VERSION < "2.6.0"
```
jaymzh, line 5: shall we leave this as well?

jaymzh (changes requested):

johnmccrae:
@marcparadise Is this still needed?

PR #399, opened 2024-10-22
Download install sh from new download api

ashiqueps:
## Description

The install.sh and install.ps1 files are moved to the new download API and maintained there. If the options has the new key, instead of building the file, download it from the new download api.

## Types of changes

## Checklist:

- [ ] If `Gemfile.lock` has changed, I have used `--conservative` to do it and included the full output in the Description above.

### DIFF
--- a/lib/mixlib/install/generator.rb
+++ b/lib/mixlib/install/generator.rb
@@ -22,11 +22,10 @@ module Mixlib
   class Install
     class Generator
       def self.install_command(options)
-        if options.for_ps1?
-          PowerShell.new(options).install_command
-        else
-          Bourne.new(options).install_command
-        end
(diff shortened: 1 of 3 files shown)


jaymzh (approved):
Thanks!

jaymzh:
After discussion, this PR isn't the cause of the failure, so we'll merge.

However, we *do* need to fix the build, and here's the context, so that it is written down somewhere:

This repo (`mixlib-install`), is the bootstrap mechanism for folks to get a modern build of ruby from a **system ruby** which may be old, and may be even an unsupported version. So, we need to look at all supported platforms, determine the oldest system ruby, and then bump our min version to that (or, that may mean keeping it where it is), then fixing our rubygem dependencies to accommodate that version.

